PR #39, opened 2022-09-16
Aim at consistent use for indexing and arrays.

jwboth:
Following the often used terms 'Cartesian indexing' and 'matrix indexing', we also include their use in daria. The first addresses xy indexing, e.g., for coordinates using a (x,y) format. The latter addresses pixels in images, using a (row,col). Since cv2 often uses a (col,row) format, we also introduce the term 'reverse matrix indexing'. From now on, we aim at using this terminology.

In addition, in this overhaul, many (almost all) pairs of two numbers are encoded as numpy arrays. Still some routines accept lists as input to ease in the initializiation of some classes. But these get converted to arrays under the hood. Arrays are often easier to handle in computations and will therefore be prioritized in many routines.

### DIFF
--- a/examples/segmentation_watershed.py
+++ b/examples/segmentation_watershed.py
@@ -13,7 +13,7 @@ from scipy import ndimage as ndi
 
 # Color image
 # TODO: provide path to baseline image
-image_color = cv2.imread("../images/fluidflower/Baseline.jpg")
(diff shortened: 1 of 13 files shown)


jwboth:
@EStorvik All examples run under the current changes and produce the same results as before.

PR #40, opened 2022-09-17
Apply transformation computed from compaction

jwboth:
This PR provides all required tools for the compaction analysis. Aside of determining the compaction map, now also transforming of images according to compaction maps is available. The transformation uses piecewise perspective transforms, such that the resulting image is continuous. Furthermore, this approach does not require any interpolation. Yet, it is slightly time consuming, and should only be considered in the post-analysis.

The tools are added in a general manner. It is the example which has the task to introduce the tools for analyzing compaction. But other than that, the tools are first and foremost tools to study translations between images and patches etc.


jwboth:
Further developments depend on the current changes. Therefore, this PR is merged without review.

PR #49, opened 2022-09-21
Tidy up examples

jwboth:
Include latest curvature correction updates in examples, unify some tests, remove obsolete tests, and add requirement to allow a notebook to successfully run.


examples/co2_and_tracer_analysis.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Returns:
+        np.ndarray: corrected image
+    """
+
+    # Define curvature correction object, initiated with config file
+    # (which can be created the workflow presented in the Jupyter notebook
+    # examples/notebooks/curvature_correction_walkthrough.ipynb).
+    curvature_correction = daria.CurvatureCorrection(
+        config_source="./images/config.json", width=2.8, height=1.5
```
jwboth, line 29:
```suggestion
        config_source="./images/config.json"
```

requirements.txt
```diff
@@ -5,4 +5,5 @@ matplotlib
 scikit-image
 future
 colour-science
-colour-checker-detection
\ No newline at end of file
+colour-checker-detection
+ipympl
```
EStorvik, line 9: I was not 100% sure if we wanted to require this, but I guess it is not too big, and it probably makes sense.

examples/co2_analysis.py
```diff
@@ -3,76 +3,97 @@
...
+    )
 
-    plt.imshow(img)
-    plt.show()
+    # Apply curvature correction.
+    img = curvature_correction(img)
 
-    # Color correction
+    # Transform to RGB space.
+    img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
```
EStorvik: I'm thinking that there should be a way to hide the change in color space within the colorcorrection. Perhaps by first creating a daria image, and then use the colorcorrection in the daria image and make use of the colorspace attribute?

EStorvik (approved):
Looks very good! I made two comments, but the only one that (slightly) matters is probably for another time. I think that we can just merge these changes.

PR #50, opened 2022-09-22
Add tools to analyze concentrations

jwboth:
The branch namen is not good.

Concentration analysis class added.


jwboth:
Have to review later... merge needed straight away.

PR #53, opened 2022-09-22
Retrieve metadata from jpg files using PIL

jwboth:
Use PIL to open images in `Image`. PIL allows also reading metadata as time stamp. This information can be very useful when considering time series.

### DIFF
--- a/examples/co2_and_tracer_analysis.py
+++ b/examples/co2_and_tracer_analysis.py
@@ -3,6 +3,8 @@ Example script for simple image analysis. By comparison of images
 of the same well test, a tracer concentration can be determined.
 """
 
+from pathlib import Path
+
 import cv2
 import matplotlib.pyplot as plt
 import numpy as np
@@ -26,7 +28,7 @@ def preprocessing(img: np.ndarray) -> np.ndarray:
     # (which can be created the workflow presented in the Jupyter notebook
     # examples/notebooks/curvature_correction_walkthrough.ipynb).
     curvature_correction = daria.CurvatureCorrection(
-        config_source="./images/config.json", width=2.8, height=1.5
+        config_source=Path("./images/config.json"), width=2.8, height=1.5
     )
 
     # Apply curvature correction.
--- a/examples/co2_and_tracer_analysis.py
+++ b/examples/co2_and_tracer_analysis.py
@@ -3,8 +3,6 @@ Example script for simple image analysis. By comparison of images
 of the same well test, a tracer concentration can be determined.
 """
 
-from pathlib import Path
-
 import cv2
 import matplotlib.pyplot as plt
 import numpy as np
@@ -28,7 +26,7 @@ def preprocessing(img: np.ndarray) -> np.ndarray:
     # (which can be created the workflow presented in the Jupyter notebook
     # examples/notebooks/curvature_correction_walkthrough.ipynb).
     curvature_correction = daria.CurvatureCorrection(
-        config_source=Path("./images/config.json"), width=2.8, height=1.5
+        config_source="./images/config.json", width=2.8, height=1.5
     )
 
     # Apply curvature correction.
@@ -39,8 +37,8 @@ def preprocessing(img: np.ndarray) -> np.ndarray:
 
     # Apply color correction - need to specify a crude ROI containing the color checker.
     roi_cc = (slice(0, 600), slice(0, 700))
-    colorcorrection = daria.ColorCorrection()
-    img = colorcorrection.adjust(img, roi_cc)
+    color_correction = daria.ColorCorrection()
+    img = color_correction(img, roi_cc)
 
     return img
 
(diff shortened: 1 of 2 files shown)


examples/co2_and_tracer_analysis.py
```diff
@@ -26,7 +28,7 @@ def preprocessing(img: np.ndarray) -> np.ndarray:
     # (which can be created the workflow presented in the Jupyter notebook
     # examples/notebooks/curvature_correction_walkthrough.ipynb).
     curvature_correction = daria.CurvatureCorrection(
-        config_source="./images/config.json", width=2.8, height=1.5
+        config_source=Path("./images/config.json"), width=2.8, height=1.5
```
jwboth, line 31: I suddenly got an error here. I don't know. You solved the issue about Path and str yesterday. Is this still following your philosphy?

jwboth: @EStorvik

EStorvik: My idea was to only pass strings, and then inside the curvature correction the Path functon would be used.

src/daria/image/image.py
```diff
@@ -78,9 +85,19 @@ def __init__(
...
         if isinstance(img, np.ndarray):
             self.img = img
+
+            # Come up with default metadata
             self.name = "Unnamed image"
+            self.timestamp = None
+
         elif isinstance(img, str):
-            self.img = cv2.imread(str(Path(img)))
+            pil_img = PIL_Image.open(img)
```
EStorvik, line 94: Should this be str(Path(img))? I am not sure if this is the best way to do things. Currently I am playing around in windows to see how this works there.

src/daria/image/image.py
```diff
@@ -175,6 +192,8 @@ def write(
             metadata_path (str): path to metadata (only folders); the metadata file
                 has the same name as the image (and a .txt ending)
         """
+        # TODO need to make sure that image is in BGR format
```
EStorvik, line 195: Can't you just make a conditional statement to self.colorspace?

EStorvik (approved):
It looks good, but we should perhaps discuss the use of Path. I'm looking into it and we can discuss tomorrow!

examples/co2_and_tracer_analysis.py
EStorvik: I now realize that you would like to send "Path"s to be read by daria? If so, that is ok with me, but I think that reading with str(Path("...")) inside the functions works equally well. We should choose one convention.

jwboth: I agree. The important thing will be to be sure, that it works both on Linux and Windows.

Users on Windows will use the backslash convention - btw. for each backslash, one needs to use two in Windows...

I will have to read more on pathlib before claiming I know what is going on.

src/daria/image/image.py
jwboth: As you could notice above, I did not get, that this was how you were foreseeing using pathlib...

src/daria/image/image.py
jwboth: Are you finished with defining a flag for colorspaces? I did not notice that. Therefore, I was just adding a todo note.

jwboth:
> It looks good, but we should perhaps discuss the use of Path. I'm looking into it and we can discuss tomorrow!

I agree. This is not very urgent at the moment anyhow.

examples/co2_and_tracer_analysis.py
EStorvik: I was using windows all yesterday and is seems that Path really does all the job for us. I even tried mixing / and \, and it always worked as long as it was inside Path.

jwboth: I will leave the responsibility for pathlib to you :) Thanks for the effort.

src/daria/image/image.py
jwboth: It turns out (in contrary to cv2) that PIL actually allows Path as input argument. So based on your comment above, I assume, something like this would be most unviersal and at the same time short?
```suggestion
            pil_img = PIL_Image.open(Path(img))
```

EStorvik: That looks very good! Let us try to have it like this whenever we use PIL then.

EStorvik:
Will you just merge this? I want to make some changes to the initialization of daria.Image (including the possibility to correct for curvature and color right away), and changing inputs to kwargs. It would be cool to start from these changes as well.

jwboth:
> Will you just merge this? I want to make some changes to the initialization of daria.Image (including the possibility to correct for curvature and color right away), and changing inputs to kwargs. It would be cool to start from these changes as well.

OK! Let me add one more very tiny PR after that, with suggestions of simplifcations in the corrections. If you aggree on them, they me be also relevant for your current effort.

PR #55, opened 2022-09-22
Some examples are now pytested to execution

EStorvik:
pytest now runs to check if all the examples are running fully. Additionally, the paths in the examples has been modified to be completely os independent with the use of the os python module.


jwboth (approved):
Nice!

PR #56, opened 2022-09-23
Simplify corrections

jwboth:
Several small changes to components of the correction routine. In particular, maybe we can decide that each main routine of a correction or analysis tool uses `__call__`. In addition, we should use more config files, and this PR is a step towards that. This will be developed further in a soon to be opened PR.


EStorvik (approved):
Looks good!

PR #57, opened 2022-09-23
Initialize image

EStorvik:
Made quite drastic changes to the initialization process of daria Images, as well as modifying all the examples accordingly. I think that it works quite well now, but there are, as always room for improvement.

Moreover, the colortags have been improved upon and colorchannel methods have been included to the image class.


src/daria/image/image.py
```diff
@@ -49,38 +50,68 @@ class Image:
...
-        metadata_path: Optional[str] = None,
-        # TODO Have to rethink the use of parameters, and possibly switch to
-        # a config file instead, that is shared among a full suite of images.
-        # It should contain information on the correction routines, necessary.
-        # It should include physical information: width, height
-        # It should include the format of timestamps included in the exifdata
+        img: Union[np.ndarray, str],
+        metadata: Optional[dict] = None,
+        curvature_correction: Optional[da.CurvatureCorrection] = None,
+        color_correction: Optional[da.ColorCorrection] = None,
```
jwboth, line 56: Again more an observation than a comment (I just have to get used to the new changes I believe) - having to pass the curvature and color correction from outside (instead of creating them inside of `daria.Image` from a config file) will require more effort from the user, in particular when working with just a single image and not a series. For the latter, it feels quite natural that the correction objects are centralized objects, as also commented on above. To me this is very much OK, but I have to say, I was surprised at first, when I saw these changes, as I suspected a decreased and increased need of user-input here.

src/daria/image/image.py
```diff
@@ -89,21 +120,25 @@ def __init__(
...
-            self.colorspace: str = "rgb"
+            self.colorspace: str = "RGB"
+            self.metadata["color_space"] = "RGB"
 
             # Read exif metadata
             self.exif = pil_img.getexif()
-            self.timestamp: datetime = datetime.strptime(
-                self.exif.get(306), "%Y:%m:%d %H:%M:%S"
-            )
+            if self.exif.get(306) is not None:
```
jwboth, line 136: Just out of curiosity (I believe this makes much sense): Did you come across an exmaple in which the previous code failed?

src/daria/image/image.py
```diff
@@ -112,25 +147,21 @@ def __init__(
...
-        else:
-            self.origo = origo if isinstance(origo, np.ndarray) else np.array(origo)
-            self.width = width
-            self.height = height
-            self.dim = dim
-        self.shape = self.img.shape
+        if curvature_correction is not None:
+            self.img = curvature_correction(self.img)
+            self.width = curvature_correction.config["crop"]["width"]
+            self.height = curvature_correction.config["crop"]["height"]
```
jwboth, line 153: Maybe it would be better, if we introduce `width` and `height` as attributes of `CurvatureCorrection`? The benefit would be that the in some sense hardcoded addressing of the navigation through the config of `CurvatureCorrection` would not be required. And it would also not be seen to the outside of `CurvatureCorrection`, which would mean the above suggestion would result in a slightly more robustness. Instead of using `width` as name for the attribute, maybe `target_width` would be a better choice, as this property is connected to the corrected image and not `CurvatureCorrection`. And as object itself, `CurvatureCorrection` does not see any image before the user does not provide it.

src/daria/image/image.py
```diff
@@ -112,25 +147,21 @@ def __init__(
...
+            self.origo = [0, 0]
```
jwboth, line 154: Do we want to overwrite `origo`? How is the procedure, if I would want to initialize an `daria.Image`, apply correction, and choose as `origo` something else then `[0,0]`? My suggestion is that the `metadata` is only updated in lines 151-155 `if metadata is None`. This would mean, `daria.Image` assumes that the provided `metadata` at initialization of a `daria.Image` is meaningful. Maybe even better, one could double check whether `width` and `height` in the `metadata` are identical to those in `curvature_correction`, and if not, a warning (maybe not an error) is thrown. And we would have to decide one convention, who is stronger: `metadata` or `curvature_correction`.

src/daria/image/image.py
```diff
@@ -112,25 +147,21 @@ def __init__(
...
         # Determine numbers of cells in each dimension and cell size
-        self.num_pixels_height, self.num_pixels_width = self.shape[:2]
+        self.num_pixels_height, self.num_pixels_width = self.img.shape[:2]
         self.dx = self.width / self.num_pixels_width
         self.dy = self.height / self.num_pixels_height
-        # ... in 3d
-        if self.dim == 3:
-            self.depth = depth
-            self.num_pixels_depth = self.shape[2]
-            self.dz = self.depth / self.num_pixels_depth
```
jwboth, line 133: I agree that we should for now consider only 2d images!

jwboth: The depth can be actually very relevant in terms of the FluidFlower. But this is very specific and should not be included here. In particular, because the depth is not constant for the FluidFlower rigs.

src/daria/image/image.py
```diff
@@ -193,26 +189,33 @@ def write(
...
         print("Image saved as: " + str(Path(path + name + file_format)))
 
-        # Write meta data
-        if save_metadata:
-            f = open(str(Path(metadata_path + name + ".txt")), "w")
-            f.write("Origo: " + str(self.origo) + "\n")
-            f.write("Width: " + str(self.width) + "\n")
-            f.write("Height: " + str(self.height) + "\n")
-            f.write("Dimension: " + str(self.dim) + "\n")
-            f.close()
```
jwboth, line 215: Agree that the management of `metadata` is not anymore business of `daria.Image`.

src/daria/image/image.py
```diff
@@ -193,26 +189,33 @@ def write(
...
+    def update_metadata(self) -> None:
```
jwboth, line 201: Not critical. Yet, one could consider using a private method here, indicated in Python by an underscore
```suggestion
    def _update_metadata(self) -> None:
```
Yet, maybe we postpone such changes to a later stage, when going over all files again just with the aim to privatize a bunch of routines.

src/daria/image/image.py
```diff
@@ -193,26 +189,33 @@ def write(
...
+    def write_metadata_to_file(self, path: str) -> None:
+        """
+        Writes the metadata dictionary to a json-file.
+
+        Arguments:
+            path (str): path to the json file
+        """
+
+        with open(str(Path(path)), "w") as outfile:
+            json.dump(self.metadata, outfile, indent=4)
```
jwboth, line 218: Oh, now I see that you wanted to keep the organization of the metadata within `Image`... Let's keep it here for now. There is no other more natural place for it right now.

src/daria/image/image.py
```diff
@@ -353,14 +360,98 @@ def copy(self) -> "Image":
         """
         Returns a copy of the image object.
         """
-        return Image(self.img, self.origo, self.width, self.height)
+        return Image(self.img, self.metadata, color_space=self.colorspace)
```
jwboth, line 363: Following my comments from above, I would rather go for removing `color_space` from the keywords, and instead move it into `metadata`. Shouldn't it anyhow be already represented in `metadata` at this stage?

src/daria/image/image.py
```diff
@@ -49,38 +50,68 @@ class Image:
...
         """
 
+        # Read metadata
+        no_colorspace_given = False
+        if metadata is not None:
+            self.metadata = metadata
+            self.width: float = self.metadata["width"]
+            self.height: float = self.metadata["height"]
+            self.origo: list[float] = self.metadata["origo"]
+            self.colorspace: str = self.metadata["color_space"]
```
jwboth, line 91: One could ask, why having the same information stored twice, as metadata is part of `Image`, storing `width`, `height`, `origo`, and `color_space`, at the same time, these are stored as attributes of `Image`. Furthermore, they are always aimed at being synchronized. Why then not only using `metadata`. Then one could remove the marked lines, and similarly lines 98-101, just below. Of course, this would require some changes in the way `metadata` is updated, but this would not be difficult either.

src/daria/image/image.py
```diff
@@ -49,38 +50,68 @@ class Image:
...
+        else:
+            self.width: float = kwargs.pop("width", 1)
+            self.height: float = kwargs.pop("height", 1)
+            self.origo: list[float] = kwargs.pop("origo", [0, 0])
+            if "color_space" in kwargs:
+                self.colorspace = kwargs["color_space"]
+            else:
+                self.colorspace = "BGR"
+                no_colorspace_given = True
+            self.colorspace: str = kwargs.pop("color_space", "BGR")
```
jwboth, line 112: Here, one would have to feed the keyword arguments to `update_metadata`.

jwboth: Or simply directly create `self.metadata = {}` and then line by line fill in `width`, etc.

src/daria/utils/conversions.py
```diff
@@ -46,9 +46,9 @@ def BGR2RED(img: da.Image) -> da.Image:
         da.Image: converged image
     """
     img_red = img.copy()
-    if img.colorspace == "bgr":
+    if img.colorspace == "BGR":
         img_red.img = img_red.img[:, :, 2]
-    elif img.colorspace == "rgb":
+    elif img.colorspace == "RGB":
         img_red.img = img_red.img[:, :, 0]
```
jwboth, line 52: Why not using simply: `img_red.toRED()` - this would in particular also change the value of `colorspace`.

src/daria/utils/conversions.py
```diff
@@ -63,7 +63,10 @@ def BGR2GREEN(img: da.Image) -> da.Image:
         da.Image: converged image
     """
     img_green = img.copy()
-    img_green.img = img_green.img[:, :, 1]
+    if img.colorspace == "BGR":
+        img_green.img = img_green.img[:, :, 1]
+    elif img.colorspace == "RGB":
+        img_green.img = img_green.img[:, :, 1]
```
jwboth, line 69: Similar as for RED.

src/daria/utils/conversions.py
```diff
@@ -77,9 +80,9 @@ def BGR2BLUE(img: da.Image) -> da.Image:
         da.Image: converged image
     """
     img_blue = img.copy()
-    if img.colorspace == "bgr":
+    if img.colorspace == "BGR":
         img_blue.img = img_blue.img[:, :, 0]
-    elif img.colorspace == "rgb":
+    elif img.colorspace == "RGB":
         img_blue.img = img_blue.img[:, :, 2]
```
jwboth, line 86: Similar as for RED.

src/daria/image/image.py
```diff
@@ -353,14 +360,98 @@ def copy(self) -> "Image":
...
+            red_img.img = red_img.img[:, :, 0]
+        elif self.colorspace == "BGR":
+            red_img.img = red_img.img[:, :, 2]
+        elif self.colorspace == "RED":
+            pass
+        else:
+            raise Exception(
+                "Only RGB or BGR images can be converted to Red at the moment"
+            )
+        red_img.colorspace = "RED"
```
jwboth, line 417: Wouldn't `R` instead of `RED` be more consistent to `RGB`? I am OK (and maybe even in favor) with using `RED` as it is simpler to search for it in the code, but thought to raise this question for possible rethinking nevertheless. Before, we switch too often conventions.

jwboth (approved):
Great effort @EStorvik. You will see I am a bit confused about the use of metadata vs keyword arguments vs also storing the same attributes in self (the latter I think I am actually less confused about). I guess it makes maybe sense, to gather all my comments on keyword arguments as input argument for `Image` before making any changes to the code. I amnot sure, if we maybe want to discuss further. In some sense I have not super strong opinion. However I think in the entire "metadata vs keyword arguments vs also storing the same attributes in self" there is duplication of concepts, which I usually would try to circumvent. I nevertheless would like to hear your comments on these things. Maybe you had serious reasons why you suggest this way.

examples/co2_and_tracer_analysis.py
```diff
@@ -86,47 +54,40 @@ def determine_tracer(img: np.ndarray, base: np.ndarray) -> np.ndarray:
...
+co2 = determine_tracer(co2_image, baseline_co2)
 
-# Plot
-plt.imshow(co2)
-plt.show(block = False)
-# Pause longer if it is desired to keep the images on the screen
-plt.pause(3)
-plt.close()
+# Plot change 3 to larger number if it is desired to keep the images on the screen
+co2.plt_show(3)
```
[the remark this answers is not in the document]
EStorvik: This was actully just introduced to make the pytest examples test conclude. There might be some way to circumvent it though.

examples/co2_and_tracer_analysis.py
```diff
@@ -86,47 +54,40 @@ def determine_tracer(img: np.ndarray, base: np.ndarray) -> np.ndarray:
...
-    return diff
+    return daria.Image(diff, base.metadata)
 
 
-# !----- Main routine for co2 analysis
+# Define curvature correction object, initiated with config file
+# (which can be created the workflow presented in the Jupyter notebook
+# examples/notebooks/curvature_correction_walkthrough.ipynb).
+curvature_correction = daria.CurvatureCorrection(
+    config_source=f"{os.path.dirname(__file__)}/images/config.json", width=2.8, height=1.5
```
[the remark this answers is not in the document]
EStorvik: I agree, we could perhaps make the string in its own variable. This is anyways just to make sure that the examples run everywhere. Any user could later include paths as they see fit for their OS.

src/daria/corrections/color/colorchecker.py
```diff
@@ -30,12 +30,13 @@ def __init__(self):
 
 
 class ColorCorrection:
-    def __init__(self):
+    def __init__(self, ROI: Optional[tuple] = None):
```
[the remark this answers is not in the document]
EStorvik: roi vs. ROI: I had no idea, ill change it back! 

I think is makes sense to have it in init. Moreover, it made it much cleanerto include the object in the __init__ of da.Image this way.

src/daria/image/image.py
EStorvik: We could do it like you say, I was just afraid that too much will be "hidden" inside the initialization of da.Image. Now, at least it is very clear that we would like to apply corrections to the images that are initialized this way. We could of course also just accept paths to config files for the corrections in the input of __init__. Then we do not have to define the correction before the Image. I am not sure what I really prefer. It is anyways only one/two lines more of user code, and it is in principle very clear what is happening at these lines.

src/daria/image/image.py
EStorvik: Yes, haha. I wanted to talk to you about this on friday.. Remind me to ask you on Tuesday, or whenever we talk again (I am in principle free all today and tomorrow as well).

src/daria/image/image.py
EStorvik: Haha, yes. But I agree. it is debatable how we would like to do all of this business.

src/daria/image/image.py
EStorvik: Oh, you are correct. This was just me confusing myself and working on the code at several places at once and changing my mind in the middle of it all. I actually don't think that the keyword argument is even read when it is written like this.

src/daria/image/image.py
EStorvik: I agree. I was just afraid that it would make it more inconvenient to extract the attributes from ouside of the object. However, I don't think we'll ever need to, and it is not really much more work...

src/daria/utils/conversions.py
EStorvik: Yes.. I actually do not really know if we even would like to keep these functions. They are now anyways a part of the Image class, and of someone wants to change the color of something externally perhaps they can just use cv2.

src/daria/image/image.py
EStorvik: Let us discuss this briefly as well, I do not have a too strong opinion, but it felt to me like "RGB" means something, while "R" alone does not. Therefore, I chose "RED". I am not sure if capital letters are the way to go in either of the places here though. What do you think?

EStorvik:
<p>I guess they are based on the "width" and "height" of the image, and are somehow automatically in the same units. Regarding the str I agree.</p>

---

 <a href='https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

EStorvik:
<p>I am not sure what happened here... This worked on my laptop, and should not require any keyboard interrupt. </p>

---

 <a href='https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

EStorvik:
<p>I think that I agree in that it should always be provided, but I am not sure whether only metadata as input should be accepted. I am open for a quick discussion on this, but I do not really now if it matters to much. Whether everything is stored as both a dictionary and separate attributes is a different thing. </p>

---

 <a href='https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

EStorvik:
<p>Haha, yes. I am actually not sure whether dx and dy are the best keywords here anyways.</p>

---

 <a href='https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

src/daria/image/image.py
```diff
@@ -49,38 +50,68 @@ class Image:
...
+                corrected at initialization.
+            kwargs (Optional arguments)
+                metadata_source (str): Path to a metadata json-file that provides
+                    metadata such as physical width, height and origo of image
+                    as well as colorspace
+                origo (np.ndarray): physical coordinates of the lower left corner
+                width (float): physical width of the image
+                height (float): physical height of the image
+                color_space (str): the colorspace of the image. So far only BGR
+                    and RGB are "valid", but more should be added at a later time.
```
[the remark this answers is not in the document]
EStorvik: We should discuss this briefly.

examples/co2_and_tracer_analysis.py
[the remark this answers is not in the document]
jwboth: Then I would add this as a comment, as currently, also the examples serve as demonstration of functionality and are therefore an introduction to daria.

examples/co2_and_tracer_analysis.py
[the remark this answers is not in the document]
jwboth: I see! Indeed, best would be something like having a flag deciding whether plots are shown with a break or without, and if the flag could be controlled from inside, but also from outside via pytest - I am not sure if something like this would be  possible...

src/daria/corrections/color/colorchecker.py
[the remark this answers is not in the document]
jwboth: I agree, please resolve after renaming `ROI`.

src/daria/image/image.py
jwboth: Agree. At this stage, I find this approach the best. Feel free to resolve.

src/daria/utils/conversions.py
jwboth: In some sense it is natural to request havign such functionality. Note that the function aims at returning a `daria.Image` and not a `np.ndarray`. Thus, cv2 may be not sufficient. Instead of having these functions, one could however consider use a flag in, e.g., `toRED()` controlling whether the current object is altered or another object is created. It could look like: `def toRED(self, return_image: bool = False) -> Optional[daria.Image]`. And if `return_image=True`, a new object is created instead of changing `self`.

src/daria/image/image.py
jwboth: Let's keep `RED`. Indeed, it is undoubted interpretable without further explanation, which is the most important. Please resolve.

src/daria/image/image.py
jwboth: > View / edit / reply to [this conversation](https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/) on ReviewNB
> 
> _jwboth commented on 2022-09-24T19:59:39Z_ ----------------------------------------------------------------
> 
> Line #2.    denoised_image.plt_show()
> Try to clean up the jupyter notebook, and not include keyboard interrupts etc. A clean file would be actually be best.
> 
> 
> _EStorvik commented on 2022-09-25T05:46:24Z_ ----------------------------------------------------------------
> 
> I am not sure what happened here... This worked on my laptop, and should not require any keyboard interrupt.

The problem lies in the nature of Jupyter notebooks, I think (also a reason why I do not like them). When executing them, you are changing them, which results in a non-empty git diff. Your local output in the terminal (including when you terminate a run) are stored in the notebook file. This is how these things got online.

jwboth: It looks like when reseting the kernel, a notebook is cleaned.

src/daria/image/image.py
jwboth: > View / edit / reply to [this conversation](https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/) on ReviewNB
> 
> _jwboth commented on 2022-09-24T19:59:39Z_ ----------------------------------------------------------------
> 
> Line #1.    grid_image = baseline_image.add_grid(dx = 10, dy = 10, color= (150,0, 100))
> Is it actually clear what units `dx` and `dy` are using? We should either define a convention (like using mm, cm, m, pixels) or we even have to include some indicator for units. We may need to provide capapbility for the conversion then as well. Adding such a possibility would possibly also allow better understanding of the input arguments.
> 
> While thinking about the arguments of `add_grid`. Would it maybe be better to also allow for `str` arguments for color and define some standard colors, which then in a utils method are converted to the correct RGB values?
> 
> 
> _EStorvik commented on 2022-09-25T05:44:59Z_ ----------------------------------------------------------------
> 
> I guess they are based on the "width" and "height" of the image, and are somehow automatically in the same units. Regarding the str I agree.

jwboth:
<p>The problem lies in the nature of Jupyter notebooks, I think (also a reason why I do not like them). When executing them, you are changing them, which results in a non-empty git diff. Your local output in the terminal (including when you terminate a run) are stored in the notebook file. This is how these things got online.</p><p> </p><p>I think reseting the kernel woudl solve the problem.</p>

---

 <a href='https://app.reviewnb.com/pmgbergen/DarIA/pull/57/discussion/'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

src/daria/image/image.py
jwboth: hmm, OK I see your point. we can discuss tomorrow. Maybe it is then safer to keep it?

src/daria/image/image.py
EStorvik: I agree that there might possibilities for improvement here. I will however not touch it now (as you said not to touch the curvature.py, and I think that the current solution is ok)

PR #59, opened 2022-09-25
Overhaul compaction analysis and unify analysis tools

jwboth:
General overhaul of `daria.CompactionAnalysis`, switching to useing `daria.Image` instead of `np.ndarray`. 

Not part of this PR, but planned when PR #57  is merged: Making `daria.ConcentrationAnalysis` using `daria.Image` and not `np.ndarray`.


jwboth:
Currently this PR does not pass the pytest. My thought is that PR #57 is merged to main before this one, allowing to merge from main back here. Based on these changes, the examples and thereby pytest can be updated accordingly.

EStorvik:
I merged #57 now. Should I review this right away?

examples/co2_and_tracer_analysis.py
```diff
@@ -126,7 +129,7 @@ def determine_tracer(img: np.ndarray, base: np.ndarray) -> np.ndarray:
 
 # Plot
 plt.imshow(tracer)
-plt.show(block = False)
+plt.show(block=False)
```
EStorvik, line 132: This file has been changed in main now. I do not have strong opinions, but it s important for pytest that the plots eventually closed themselves, unless we can get a pytest-flag of some sort

examples/compaction_analysis.py
```diff
@@ -59,8 +55,10 @@
     "in meters": True,
```
EStorvik, line 55: Slightly unrelated to this particular instance, but if we would like to go for units i am not sure if boleans are the way to go.

src/daria/analysis/compactionanalysis.py
```diff
@@ -0,0 +1,69 @@
...
+            plot (bool): flag controlling whether the deformation is also
+                visualized as vector field.
+            reverse (bool): flag whether the translation is understood as from the
+                test image to the baseline image, or reversed. The default is the
+                former one.
+        """
+        transformed_img = self.translation_analysis(img)
+        if plot:
+            self.translation_analysis.plot_translation(reverse)
+        return transformed_img
```
EStorvik, line 69: I really like the style of this!

EStorvik (approved):
Many of these files are ones that I do not know too much about, but I like the style of it all. I guess you plan to have some examples that highligt how these classes are meant to work as well? the compaction_analysis already somehow are there, but the concentration one I am not sure. Is this supposed to be incorporated into the co2_and_tracer_analysis file?

examples/compaction_analysis.py
jwboth: No they are not. This flag is in fact not important anymore as the code is not affected by it. I think what was more important than the unit was that width and height were provided in metric lengths and not pixels. But this does not matter anymore.

jwboth:
> Many of these files are ones that I do not know too much about, but I like the style of it all. I guess you plan to have some examples that highligt how these classes are meant to work as well? the compaction_analysis already somehow are there, but the concentration one I am not sure. Is this supposed to be incorporated into the co2_and_tracer_analysis file?

@EStorvik Yes, however, as `CompactionAnalysis` is using `daria.Image` and `ConcentrationAnalysis` is not, I wanted to postpone this. But this is exactly the idea for the next PR. PR #57  has been also very important before one could start working on that.

PR #61, opened 2022-09-27
Restructure and accelerate curvature correction.

jwboth:
Two main changes have been performed for `CurvatureCorrection`. First, routines have been sorted as I/O, auxiliary computations, main, etc.

Furthermore, `__call__` uses chached computations, finally requireing merely mapping coordinates only once per call. This accelerates corrections for those image, which use a series of tailored bulge, stretch etc. corrections

### DIFF
--- a/src/daria/corrections/shape/curvature.py
+++ b/src/daria/corrections/shape/curvature.py
@@ -16,6 +16,7 @@ import numpy as np
 from scipy.ndimage import map_coordinates
 
 import daria as da
+import skimage
 
 
 class CurvatureCorrection:
@@ -83,6 +84,7 @@ class CurvatureCorrection:
                     "Invalid image data. Provide either a path to an image or an image array."
                 )
             self.current_image = np.copy(self.reference_image)
+            self.dtype = self.current_image.dtype
             self.Ny, self.Nx = self.reference_image.shape[:2]
             self.in_meters = kwargs.pop("in_meters", True)
             self.width = kwargs.pop("width", 1.0)
@@ -146,7 +148,7 @@ class CurvatureCorrection:
 
     @property
     def temporary_image(self):
-        return self.current_image.astype(np.uint8)
+        return skimage.util.img_as_ubyte(self.current_image)
 
     # ! ---- Wrappers for single transformations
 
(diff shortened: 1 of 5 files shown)


src/daria/corrections/shape/curvature.py
```diff
@@ -473,3 +349,227 @@ def compute_stretch(self, **kwargs):
...
+        """
+        # FIXME: I suggest to remove the possibility of calling images via str outside
+        # of daria.Image. Opening an image comes with a pre-defined choice of the
+        # colorspace. This control is lost if we do not further pass this information.
+        # Also, modular thinking has had its success in software development for a reason.
+        # I would also say, either the user is loading the image him/herself but then
+        # is required to provide all other necessary information. Other than that, I am
+        # in favor of centralizing opening images from file. This be either in daria.Image
+        # or if need in utils (in addition). But I think, only daria.Image is the right
+        # place for it.
```
EStorvik, line 380: I agree, but then we need to find a good and easy way of "making" the correction. Oncle we have the correction config file, everything else is taken care of by daria.Image anyways.

jwboth: OK, then I suggest, I add a comment in the documentation here that the option of providing a path should be used with care etc.

EStorvik (approved):
Looks very good! Is there a nice way of testing the changes in github? Or should I then pull the branch and test everything locally? I guess you have tested that everything works as expected anyways!

src/daria/corrections/shape/curvature.py
EStorvik: Ok, that sounds good!

jwboth:
> Looks very good! Is there a nice way of testing the changes in github? Or should I then pull the branch and test everything locally? I guess you have tested that everything works as expected anyways!

Yes, I have tested only locally. Indeed a test would be great... but then we would need the previous code as well for testing that both do the same job. I am quite sure they do. But one can never know.

EStorvik:
What is the best workflow for these PRs? Should I now merge? Or do you want to do something more?

jwboth:
> What is the best workflow for these PRs? Should I now merge? Or do you want to do something more?

I am not sure. I realized, when running the curvature notebook, that the plotting does not always work. And the problem is that there is conversions missing. I am not sure whether these have been introduced through this code. However, float images are required to be in the range of [0,1]. Sometimes daria produces however float images through corrections, which are in the range [0,255]. Then one has to convert back in order to see the image properly. This is anyhow a deeper issue. Shall we wait with merging until this issue is resolved?

jwboth:
@EStorvik I thought it is maybe best if we ensure robustness instead of try to hack something that sometimes works and sometimes not. This will lead to some format conversions, which on the one hand may alsways take some time, and may also remove information.

However, the only situation, where potentially critical information is removed is when routines obtain images of dtype np.uint8, i.e. in the range of [0,255], and when they return np.float32 in the range of [0,255]. Switching back to np.uint8 hence removes fairly litte information, but still. I know.

PR #65, opened 2022-09-30
Overhaul of analysis tools and correpsonding examples

jwboth:
Overview of changes in this PR:
- Some small changes making existing routines compatible with a larger range of input arguments (e.g., `np.ndarray` as well as `daria.Image`, or `str` as well as `pathlib.Path`).
- Add capabilities to generate new config files from existing ones if a resized version of original images is considered.
- Add calibration capabilities to the concentration analysis. 
- Examples are up to date.


jwboth:
NB: Merging without external review. Though the changes have been tested on a current project. They seem to do what they have been intended for.

PR #66, opened 2022-10-04
Hotfix: Show the image in curvature correction.

jwboth:
The method `show_image` was lacking a `plt.show()` to in fact show the image.


jwboth:
As simple and hotfix, this is accepted without further review.

PR #67, opened 2022-10-04
Hotfix: Plot image in RGB in curvature correction.

jwboth:
Images read with PIL are coded in RGB. The conversion between BGR and RGB in the plotting function of the curvature correction is obsolete and has been removed. Images are now plotted in the expected color spectrum.


EStorvik (approved):
A very reasonable change.

PR #69, opened 2022-10-05
Hotfix: Fix the copy of Image

jwboth:
The method `copy` of `Image` has been returning a shallow copy. Though a deepcopy has been intended. This is fixed with the current PR.


EStorvik (approved):
Looks very good, I did not think of it, but it makes sense to be this careful.

PR #70, opened 2022-10-06
Origin/hotfix windows curvcorrection

EStorvik:
Minor changes so that round is used instead for int some places, and dtype = np.int64 is forced (also on windows where np.int32 is default) in curvature correction which fixes the rounding problems for large images.


jwboth (approved):
Looks good to me. Thanks for the effort!

PR #72, opened 2022-10-10
Evaluation of compaction in arbitrary points

jwboth:
Bugfix for computing the origo when restricting a `daria.Image` using ROIs.

Add capability to evaluate compaction in arbitrary points. Both pixel and metric units can be used, though the latter is the standard. The use of the evaluation is showcased in the compaction analysis example.


EStorvik (approved):
Looks very good!

PR #73, opened 2022-10-14
Feature/compare segmentations

EStorvik:
Comparison of different segmented images. There are still some flaws in relation to generality (at least for increasing number of components in the segmented images), and it might be that the way the gray colors are implemented should be changed (perhaps gray is not even the correct to use, and different other color spaces would be better).  For now, however, I think that it works sufficiently well, and it should be tested on more segmentations.


src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+            colors (np.ndarray): Array of different colors that should
+                indicate uniqe components in each segmentation.
+            light_scaling (float): Indicate how much lighter the second
+                component should be scaled in its unique color.
+    """
+    # Define components
+    components: list = kwargs.pop("components", [1, 2])
+
+    # Define number of segmentations
+    N: int = len(segmentations)
```
jwboth, line 29: Use more descriptive variable name instead of single character variables for better search-functionality.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+
+            # Overlap of second component
+            return_image[
+                np.where(
+                    np.logical_and(
+                        segmentations[k] == components[1],
+                        segmentations[i] == components[1],
+                    )
+                )
+            ] += gray_base[1]
```
jwboth, line 91: Use a simple for loop over number of components here to shorten the code.

jwboth: np.where should be obsolete here as well.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+
+            # Overlap of different components
+            return_image[
+                np.where(
```
jwboth, line 95: Again, np.where should be obsolete here, I believe.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+                        ),
+                        np.logical_and(
+                            segmentations[k] == components[1],
+                            segmentations[i] == components[0],
+                        ),
+                    )
+                )
+            ] += gray_base[2]
+
+    # Determine locations (and make modifications to return image) of unique coomponents
```
jwboth, line 109:
```suggestion
    # Determine locations (and make modifications to return image) of unique components
```

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Assert that there are a sufficient amount of colors
+    # and that all of the segmentations are of equal size
+    assert len(colors[:, 0, 0]) >= N
+    for i in range(N - 1):
+        assert segmentations[0].shape == segmentations[i + 1].shape
+
+    # Create the return array
+    return_image: np.ndarray = np.full(
+        shape=segmentations[0].shape + (3,), fill_value=[0, 0, 0], dtype=np.uint8
```
jwboth, line 66:
```suggestion
    return_image: np.ndarray = np.zeros(
        segmentations[0].shape[:2] + (3,), dtype=np.uint8
```

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+    np.clip(colors_light, 0, 255, out=colors_light)
+    colors: np.ndarray = np.hstack((colors, colors_light))
+
+    # Assert that there are a sufficient amount of colors
+    # and that all of the segmentations are of equal size
+    assert len(colors[:, 0, 0]) >= N
+    for i in range(N - 1):
+        assert segmentations[0].shape == segmentations[i + 1].shape
+
+    # Create the return array
```
jwboth, line 64:
```suggestion
    # Create the return array, prepare for colored image with RGB colors
```

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+    light_scaling: float = kwargs.pop("light_scaling", 1.5)
+    colors_light: np.ndarray = np.trunc(light_scaling * colors)
+    np.clip(colors_light, 0, 255, out=colors_light)
+    colors: np.ndarray = np.hstack((colors, colors_light))
+
+    # Assert that there are a sufficient amount of colors
+    # and that all of the segmentations are of equal size
+    assert len(colors[:, 0, 0]) >= N
+    for i in range(N - 1):
+        assert segmentations[0].shape == segmentations[i + 1].shape
```
jwboth, line 62:
```suggestion
    for i in range(1,N):
        assert segmentations[0].shape == segmentations[i].shape
```

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+        ),
+    )
+    light_scaling: float = kwargs.pop("light_scaling", 1.5)
+    colors_light: np.ndarray = np.trunc(light_scaling * colors)
+    np.clip(colors_light, 0, 255, out=colors_light)
+    colors: np.ndarray = np.hstack((colors, colors_light))
+
+    # Assert that there are a sufficient amount of colors
+    # and that all of the segmentations are of equal size
+    assert len(colors[:, 0, 0]) >= N
```
jwboth, line 60:
```suggestion
    assert colors.shape[0] >= N
```

Why not "=="?

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+                to be provided.
+            gray_colors (np.ndarray): array of three different scales of
+                gray, one for each of the different combinations of
+                components in the segmentations.
+            colors (np.ndarray): Array of different colors that should
+                indicate uniqe components in each segmentation.
+            light_scaling (float): Indicate how much lighter the second
+                component should be scaled in its unique color.
+    """
+    # Define components
```
jwboth, line 25: I would be more precise and say something like that these are the considered components. The segmentations are allowed to have values outside this list.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+import numpy as np
+
+
+def compare_segmentations(*segmentations, **kwargs) -> np.ndarray:
+    """
+    Comparison of segmentations.
+
+    Args:
+        segmentations (asterisk argument): Allows to provide an arbitraty
+                number of segmented images to be compared
```
jwboth, line 12: I would rather say numpy arrays instead of images, jpg is not allowed e.g. Also mention that these arrays are expected to be filled with int values.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+        optional keyword arguments (kwargs):
+            components (list): list of the different components that are
+                found in the segmented images. So far only two are allowed
+                to be provided.
+            gray_colors (np.ndarray): array of three different scales of
+                gray, one for each of the different combinations of
+                components in the segmentations.
```
jwboth, line 19: Mention that these have to be provided in RGB space (and not in grayscale space), as long it is implemented as right now.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+                    )
+                )
+            ] += gray_base[2]
+
+    # Determine locations (and make modifications to return image) of unique coomponents
+    for c in components:
+        for k in range(N):
+            only_tmp: np.ndarray = segmentations[k] == c
+            for j in filter(lambda j: j != k, range(N)):
+                only_tmp = np.logical_and(only_tmp, segmentations[j] == 0)
```
jwboth, line 114: I suggest adding another keyword argument which allows to provide the "default value" or non-active component, which here is 0, but it does not have to. This information is hidden at the moment.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -0,0 +1,118 @@
...
+
+            return_image[np.where(only_tmp)] = colors[k, components.index(c)]
```
jwboth, line 116: Try without np.where here as well. And I would usually use `enumerate` here instead of `index`. Since the number of components won't be large, the cost will be bounded here. So it is not critical, but I think it would be the more standard way of doing this.

jwboth (approved):
Looks good for now. See some comments.

src/daria/analysis/segmentationcomparison.py
EStorvik: Done

src/daria/analysis/segmentationcomparison.py
EStorvik: Both worked well, I didnt know that where was obsolete

src/daria/analysis/segmentationcomparison.py
EStorvik: I changed the wording in "components"

src/daria/analysis/segmentationcomparison.py
EStorvik: I actually thought about this, but aparently I forgot to add it myself. Thanks!

src/daria/analysis/segmentationcomparison.py
EStorvik: Smart. Added it now

src/daria/analysis/segmentationcomparison.py
EStorvik: I gave 6 colors in the default, and it doesn't really matter if we have more colors that unique components as only the first ones will be used anyways.

jwboth: Seems a bit random but OK for now. I think providing default list of standard colors (green, red, etc.) and extracting N many of them would feel like the most natural approach. Or one could utilize asuitable  linear colormap from matplotlib and simply consider N many equally distanced colors.

EStorvik: That was a much better idea, and I fixed it now.

PR #74, opened 2022-10-15
Add new machine-learning free color correction

jwboth:
It turns out, the color correction detection from colour science sometimes has troubles with detecting half colorcheckers (although these are the only ones used in the calibration). To mitigate this a new color correction routine has been added, in which the ROI corresponding to the color checker has to be specified by providing the pixel coordinates of the white L marks on the color checker. The remaining routine follows the same idea, however is much simpler, as it does not utilize a EOTF transformation, but only works in linear RGB space. 

The example has been extended to showcase the use of the new color correction routine. The results after all are slightly different. There are several reasons to this: The lack of use of EOTF. And in fact, the machine learning based algorithms detect colours with RGB values outside the valid range of values in [0,1]. So there is quite some discrepancy, which is also visible in the final result. My current feeling is therefore, one may have to be a bit careful with the use of the automatic color correction. Instead one should possibly use the manual one. Setting it up is not much more tedious then defining a coarse ROI continaing the colorchecker.


EStorvik (approved):
I think it looks good, but it seems slightly strange to have two colorcorrections. Is your sense that the manual one simply is better than the old one? It looks like they both need the same type of input to work properly? Is it situational which one works better? We should probably give some indication as to which one should be used when. 

I would always use the non-manual one just from the name of them, because it seems more likely that I would make a mistake myself in using a manual one than an automatic one.

jwboth:
> I think it looks good, but it seems slightly strange to have two colorcorrections. Is your sense that the manual one simply is better than the old one? It looks like they both need the same type of input to work properly? Is it situational which one works better? We should probably give some indication as to which one should be used when.
> 
> I would always use the non-manual one just from the name of them, because it seems more likely that I would make a mistake myself in using a manual one than an automatic one.

I agree with you. Two do not look to good. I think experience will have to show. This is the thing with images. Automatic things may not be robust on all inputs. And the currently automatic one fails. It throws an error that it does not even find the color checker sometimes. And I am actually also not too happy with the performance of the color matching routine. It goes via EOTF (nonlinear RGB space), and through these mappings often colors are outside the valid range [0, 255].

Setup for the manual is slightly dfferent. It requires four coordinates instead of a ROI provided as tuple of slices. You choose what is more straightforward for non-experts ;)

jwboth:
Let's have a short discussion on naming etc. before merging.

src/daria/corrections/color/manual_colorcorrection.py
```diff
@@ -0,0 +1,255 @@
...
+            ]
+        ).astype(np.float32)
+
+        # Convert reference swatch colors to RGB
+        self.reference_swatches_rgb = np.squeeze(
+            cv2.cvtColor(np.atleast_3d(reference_swatches_lab), cv2.COLOR_LAB2RGB)
+        )
+
+
+class ManualColorCorrection:
```
jwboth, line 61: Let's have a discussion regarding names etc. before merging.

PR #75, opened 2022-10-16
Fix of several issues in CurvatureCorrection

jwboth:
Several sloppy issues have been detected in CurvatureCorrection:
- Wrong precomputation of cached transformed pixels for fast evaluation of the curvature correction.
- Use of integers for pixel coordinates instead of floats, resulting in rounding when computing the correction. These values are not required to be of type integer as they will anyhow merely be used in the interpolation. When using integers, light discontinuities could be observed for transformations.
- Wrong use of reference dimensions in the computation of bulge and stretch factors.
- Lack of usability of curvature correction for grayscale images.

These issues have been solved in this PR, and tested by comparison with previous versions of the CurvatureCorrection. Note the name of the branch indicates loss of permformance. However, the final changes do not alter the performance of the code.

In addition (although not perfectly matching into this PR), argument lists have been extended to images for some auxiliary functions.

### DIFF
--- a/examples/images/config.json
+++ b/examples/images/config.json
@@ -31,13 +31,13 @@
     "bulge": {
         "horizontal_bulge": -0.0,
         "horizontal_center_offset": 0,
-        "vertical_bulge": -2.5945990825497644e-08,
-        "vertical_center_offset": -33
+        "vertical_bulge": -2.7297344514325645e-08,
+        "vertical_center_offset": -31
     },
     "stretch": {
         "horizontal_stretch": -1.3307656210674836e-08,
         "horizontal_center_offset": -114,
         "vertical_stretch": -0.0,
-        "vertical_center_offset": 126
+        "vertical_center_offset": 168
     }
 }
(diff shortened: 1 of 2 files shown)


EStorvik (approved):
Looks again good! I'm just a bit confused about the 

"Wrong use of reference dimensions in the computation of bulge and stretch factors." comment. what does that mean? and where have you applied it?

jwboth:
> Looks again good! I'm just a bit confused about the
> 
> "Wrong use of reference dimensions in the computation of bulge and stretch factors." comment. what does that mean? and where have you applied it?

in the auxiliary methods as `compute_bulge` and `compute_stretch` previously `self.Nx` and `self.Ny` have been used to address the dimensions of the image. However, when cropping an image, the image size actually changes, while the dimensions refer to the input image. Therefore, finally the returned parameters from the auxiliary methods are slightly off. The effect is not too big, but it exists in terms of center offset and bulge/stretch factor.

PR #76, opened 2022-10-17
Add drift correction

jwboth:
When analyzing a series of images, they ought be aligned wrt a fixed ROI. This has been so far possible with DarIA. However, it has not been integrated in the standard workflow when constructing for instance a `daria.Image`. Therefore, a new correction object has been introduced, automatizing the correction due to translation/drift (the latter name is suggested, as there exist already two translation objects in darIA).  

In addition, the correction has been integrated in `daria.Image` as well as in the `compaction_analysis.py` example, in which previously the `daria.TranslationEstimator` has been used to align images. 

Note: Aligning images is also relevant when working with concentration analysis, where images are directly compared.

### DIFF
--- a/src/daria/image/image.py
+++ b/src/daria/image/image.py
@@ -51,8 +51,9 @@ class Image:
         self,
         img: Union[np.ndarray, str, Path],
         metadata: Optional[dict] = None,
-        curvature_correction: Optional[da.CurvatureCorrection] = None,
+        drift_correction: Optional[da.DriftCorrection] = None,
         color_correction: Optional[da.ColorCorrection] = None,
+        curvature_correction: Optional[da.CurvatureCorrection] = None,
         **kwargs,
     ) -> None:
         """Constructor of Image object.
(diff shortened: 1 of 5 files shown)


src/daria/image/image.py
```diff
@@ -65,12 +66,13 @@ def __init__(
         Arguments:
             img (Union[np.ndarray, str, Path]): image array with matrix indexing
             metadata (dict, Optional): metadata dictionary, default is None.
-            curvature_correction (daria.CurvatureCorrection, Optional):
-                Curvature correction object. Default is none, but should be included
-                if the image is to be curvature corrected at initialization
+            drift_correction (daria.DriftCorrection, optional): Drift correction object.
```
EStorvik, line 69:
```suggestion
            drift_correction (daria.DriftCorrection, Optional): Drift correction object.
```

EStorvik (approved):
Seems like a very convenient way to do things! Great!

src/daria/image/image.py
EStorvik: I guess that this doesnt even matter... Just merge when you would like to!

PR #77, opened 2022-10-18
Update examples and use the new color correction in all of them

jwboth:
The tracer example did not procude nice images anymore due to wrong use of the new color checker. This has been fixed now.


EStorvik (approved):
Good!

PR #78, opened 2022-10-19
Determine bounding box from coordinates

jwboth:
`daria.DriftCorrection` requires a tuple of slices as ROI, while `daria.ColorCorrection` works best for ROIs provided by providing a set of coordinates to define a quadrilateral. This PR adds a simple util function, which computes the bounding box for a set of coorinates, returned as tuple of slices. Thus, this can be used to convert from one to the other. This functionality is integrated in `daria.DriftCorrection` to mitigate the need for defining two different ROIs when possibly wanting to identifying a color palette as fixed point.

### DIFF
--- a/src/daria/utils/box.py
+++ b/src/daria/utils/box.py
@@ -1,7 +1,7 @@
 import numpy as np
 
 
-def bounding_box(coords: np.array) -> tuple:
+def bounding_box(coords: np.ndarray) -> tuple:
     """
     Determine bounding box for a set of given coordinates.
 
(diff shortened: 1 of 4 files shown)


src/daria/utils/box.py
```diff
@@ -0,0 +1,22 @@
+import numpy as np
+
+
+def bounding_box(coords: np.array) -> tuple:
```
jwboth, line 4:
```suggestion
def bounding_box(coords: np.ndarray) -> tuple:
```

EStorvik (approved):
Very neat functionality to have!

PR #79, opened 2022-10-19
uint16 compatability

EStorvik:
Modified in places where it seemed natural in order to make uint16 images feasible. both color correction and curvature correction seem to work fine also for the uint16 type images as it is now.

Note:
- Experimental color correction still only allows uint8


src/daria/corrections/shape/curvature.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, config: Optional[dict] = None, **kwargs) -> None:
                 self.reference_image = im_source
 
             elif isinstance(im_source, str):
-                self.reference_image = np.array(PIL_Image.open(Path(im_source)))
+                self.reference_image = cv2.imread(im_source, cv2.IMREAD_UNCHANGED)
```
jwboth, line 82: Shall we also convert to RGB here to not introduce any visual changes to the jupyter notebook on curvature correction?

src/daria/corrections/shape/curvature.py
```diff
@@ -423,7 +426,7 @@ def __call__(
 
         # Read image
         if isinstance(img, str) or isinstance(img, Path):
-            img = np.array(PIL_Image.open(Path(img)))
+            img = cv2.imread(img, cv2.IMREAD_UNCHANGED)
```
jwboth, line 429:
```suggestion
            img = cv2.imread(str(img), cv2.IMREAD_UNCHANGED)
```
cv2 does not accept `Path` unfortunately. Again BGR -> RGB?

src/daria/image/image.py
```diff
@@ -205,29 +212,46 @@ def colorspace(self) -> str:
...
+                [
+                    cv2.IMWRITE_TIFF_COMPRESSION,
+                    1,
+                    cv2.IMWRITE_TIFF_XDPI,
+                    350,
+                    cv2.IMWRITE_TIFF_YDPI,
+                    350,
+                ],
+            )
+        else:
```
jwboth, line 251: I will have to trust you on that part.

jwboth:
`daria.Patches` includes a transformation to ubyte. Have you checked that file? Also the regularization.

jwboth:
Other than that, it looks OK to me. Difficult to say actually. It is not what you did, but rather, not knowing what you possibly missed (although I also cannot think of anything else), that makes me nervous. But I guess practice will show!

After all, I am surprised how little hardcoded daria was for uint8. I am however still afraid some of the skimage routines may be unhappy. One of the routines which throw a warning on my computer are the skimage.filters.rank routines. We will see :)

jwboth (approved):

PR #80, opened 2022-10-20
Hotfix: Image.write

EStorvik:
Made several very minor changes, but there was a bug in the image write, that did not take into account changes made to the image outside of daria.Image, so that it would become wrong when writing a float type image. This is fixed now

### DIFF
--- a/src/daria/corrections/color/colorcorrection.py
+++ b/src/daria/corrections/color/colorcorrection.py
@@ -65,6 +65,9 @@ class ColorCorrection:
     X-rite color checker has four corner landmarks on the colorchecker. The pixel
     coordinates of these have to be provided as input parameters, starting with the
     corner close to the brown swatch, and continuing in counter-clockwise direction.
+    The color correction is applied prior to curvature correction in 
+    the daria.Image initialization, and therefore, the roi should be with respect to
+    the uncorrected image.
     """
 
     def __init__(
--- a/src/daria/corrections/color/colorcorrection.py
+++ b/src/daria/corrections/color/colorcorrection.py
@@ -118,10 +118,10 @@ class ColorCorrection:
         """
         # Make sure that the image is in uint8 or uint16 format
         if image.dtype in [np.uint8, np.uint16]:
-            image = skimage.util.img_as_float(image)
+            image = skimage.img_as_float(image)
         if image.dtype not in [np.float32, np.float64]:
             raise ValueError(
-                "Provide image in np.uint8, np.float32, or np.float64 format."
+                "Provide image in np.uint8, np.uint16, np.float32, or np.float64 format."
             )
 
         # Extract part of the image containing a color checker.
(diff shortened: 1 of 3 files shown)


src/daria/corrections/color/colorcorrection.py
```diff
@@ -65,6 +65,9 @@ class ColorCorrection:
     X-rite color checker has four corner landmarks on the colorchecker. The pixel
     coordinates of these have to be provided as input parameters, starting with the
     corner close to the brown swatch, and continuing in counter-clockwise direction.
+    The color correction is applied prior to curvature correction in
+    the daria.Image initialization, and therefore, the roi should be with respect to
+    the uncorrected image.
```
jwboth, line 70: I chose this order once, because curvature correction did crop important parts of a color palette in one example. So it made more sense at that time. Now we changed the curvature correction routine. What do you think? Shall we do it differently? Or shall we instead include finding the ROI for the color correction as first step in the jupyer notebook and make it about setting up the correction routines?

src/daria/corrections/color/colorcorrection.py
```diff
@@ -115,22 +118,21 @@ def __call__(
         """
         # Make sure that the image is in uint8 or uint16 format
         if image.dtype in [np.uint8, np.uint16]:
-            image = skimage.util.img_as_float(image)
+            image = skimage.img_as_float(image)
```
jwboth, line 121: I am assuming the methods are identical, or?

jwboth (approved):
Looks very good to me!

src/daria/corrections/color/colorcorrection.py
EStorvik: That sounds like a good idea!

src/daria/corrections/color/colorcorrection.py
EStorvik: I think they are the same (at least everything looks the same), but for some reason python complains that the .util part is old and not supposed to be used anymore..

src/daria/corrections/color/colorcorrection.py
EStorvik: Then we could probably also include the option for the color correction to read the config file and include the roi there.

jwboth: I fully agree, and would like that. Witb the previous version it seemed easier with not including it as json did not like slice. But nested lists (what would be needed now) work, they just have to be converted to an array after reading.

src/daria/corrections/color/colorcorrection.py
jwboth: Oh interesting! Good to know.

src/daria/corrections/color/colorcorrection.py
EStorvik: I'll fix this today then!

PR #82, opened 2022-10-20
Extend concentration analysis

jwboth:
Add more features to `daria.ConcentrationAnalysis` inculding automatic setup of cleaning filters, and adding more possibilities for choosing monochromatic color spaces.

In addition, a new class for binary segmentation is added: `daria.BinaryConcentrationAnalysis`. It is defines by inheritance and adds extensive postprocessing possibilties to tune thresholding mechanisms for segmenting images after comparison to some baseline image. This class for instance allows to distinguish water from CO2(g) and CO2 saturated water in the FluidFlower analysis.


src/daria/analysis/concentrationanalysis.py
```diff
@@ -187,11 +193,29 @@ def _extract_scalar_information(self, img: daria.Image) -> None:
...
+        """
+        Make a mono-colored image from potentially multi-colored image.
+
+        Args:
+            img (np.ndarray): image
+
+        Returns:
+            np.ndarray: monochromatic reduction of the array
+        """
+        return img
```
EStorvik, line 217: This is a bit confusing to me. It looks like a function that should to something to img and return (potentially) nothing, yet it does nothing and returns img.

src/daria/analysis/concentrationanalysis.py
```diff
@@ -216,24 +254,33 @@ def find_cleaning_filter(
             baseline_images (list of daria.Image): series of baseline_images.
             reset (bool): flag whether the cleaning filter shall be reset.
         """
+        print("Setting up a cleaning filter.")
```
EStorvik, line 257: Should this be within some verbosity check? I guess it is not needed, but I always find it nice to be able to turn of printing.

src/daria/analysis/concentrationanalysis.py
```diff
@@ -216,24 +254,33 @@ def find_cleaning_filter(
...
             self._extract_scalar_information(probe_img)
 
             # Take (unsigned) difference
             diff = skimage.util.compare_images(
                 probe_img.img, self.base.img, method="diff"
             )
 
+            # Extract mono-colored version (should be only active in case of a
+            # multichromatic comparison
+            monochromatic_diff = self._extract_scalar_information_after(diff)
```
EStorvik, line 278: why not just set it equal to diff? It looks like the same action? is it a memory thing?

src/daria/analysis/concentrationanalysis.py
```diff
@@ -216,24 +254,33 @@ def find_cleaning_filter(
...
+
             # Consider elementwise max
-            self.threshold = np.maximum(self.threshold, diff)
+            self.threshold = np.maximum(self.threshold, monochromatic_diff)
+
+        print("Finished setting up a cleaning filter.")
```
EStorvik, line 283: In case you want a verbosity flag. Not supernecessary for me.

src/daria/analysis/concentrationanalysis.py
```diff
@@ -348,3 +395,383 @@ def _determine_total_volume(self, concentration: daria.Image) -> float:
...
+        """
+        Pass the signal, assuming it identifies a binary concentration.
+
+        Args:
+            signal (np.ndarray): signal array.
+
+        Returns:
+            np.ndarray: signal array.
+        """
+        return signal
```
EStorvik, line 777: Im guess that this method also does something... Perhaps you could tell me what?

src/daria/image/image.py
```diff
@@ -573,3 +615,72 @@ def toGreen(self, return_image: bool = False) -> Optional[da.Image]:
...
+
+        Args:
+            return_image (bool): flag controlling whether the converted image
+                is returned, or converted internally.
+
+        Returns:
+            daria.Image (optional): converted image, if requested
+        """
+        # Only for RGB images for now.
+        assert self.metadata["color_space"] == "RGB"
```
EStorvik, line 631: I guess one could add a toRGB(), or similar, to allow for at least BGR images also. But one could of course do this outside as well.

src/daria/utils/box.py
```diff
@@ -2,6 +2,7 @@
 
 
 def bounding_box(coords: np.ndarray) -> tuple:
+
```
EStorvik, line 5: Why is this empty line here?

EStorvik (approved):
Super cool! I was a bit confused as to what some of these method functions actually do..? But I'm sure there is a good explanation. Would love to hear it! 

Also, I'm just assuming that the skimage denosing methods are always faster? Is there some reason to use the functionality in our denosing algorithm which permits spatially varying mu? Perhaps together with a segmentation of the sand layers?

src/daria/analysis/concentrationanalysis.py
jwboth: The main reason is to allow modular deisgn such that with inheritance this class can be very quickly tailored to a sepcific scenario. For example: I have played a bit with this design to restrict images to the third component of the HSV color space, but only used signals which have hue in a certain interval. Instead of providing these rather custom functions, inheritance felt much simpler.

src/daria/analysis/concentrationanalysis.py
jwboth: I can also remove them. Setting up a cleaning filter takes quite some time (sometimes minutes). I will adjust such that the printing only happens when it takes long time.

src/daria/analysis/concentrationanalysis.py
jwboth: Again modularity. I found it to be a differnce when the signal is restricted from tri- to monochromatic signals. Either before or after the difference. In order to allow tailoring to any given image, this modularity has been introduced.

src/daria/utils/box.py
jwboth: black

src/daria/image/image.py
jwboth: I did not want to make it too complicated at this stage, but I thought something similar.

src/daria/analysis/concentrationanalysis.py
jwboth: This is a preparation for allowing to rescale signals depending on the sandlayers. Such a conversion is included in the parent class and scales according to some calibration. The binary concentration class does not contain a calirbation yet and therefore the parent class function is overwritten here.

EStorvik: Aah, I see.

src/daria/analysis/concentrationanalysis.py
EStorvik: Sounds good!

PR #84, opened 2022-10-21
Warning when computing stretch correction

EStorvik:
Made some modification to the compute_stretch method so that it checks whether 0 stretch should be applied and gives a warning when one tries to stretch in a direction but are "dangerously close" (within 10% of total pixels in image) to the given stretch center.


src/daria/corrections/shape/curvature.py
```diff
@@ -387,18 +388,46 @@ def compute_stretch(self, img: Optional[np.ndarray] = None, **kwargs):
...
+        if (pt_dst[0] - pt_src[0]) == 0:
+            horizontal_stretch = 0
+        # Check whether point is chosen too close to the center
+        # (within 10% of total pixels), and make warning
+        elif abs(pt_src[0] - stretch_center[0]) < round(0.1 * Nx):
+            horizontal_stretch = 0
+            warn(
+                "point_source chosen too close to stretch center for correction"
+                " in horizontal direction. Please choose a value towards the corners."
+                " Horizontal stretch is set to zero now."
```
jwboth, line 406: I think it is best to mention the 10% in the warning. It is actually important that the value is also not too close to the corners, as then `Nx - pt_src[0]` or `pt_src[0]' also approache 0. So in principle, one should also check for these in the same condition. As we set up the stretch correction, the best would be to choose some value in the center of the four respective quadrants if this makes sense. So my suggestion would be to amend this to the check, as well as be more precise in the warning, to allow the user to make an adjustment without the need to guess.

src/daria/corrections/shape/curvature.py
```diff
@@ -387,18 +388,46 @@ def compute_stretch(self, img: Optional[np.ndarray] = None, **kwargs):
...
+            )
+        else:
+            # Compute the tuning parameter as explained in the notes
```
jwboth, line 409:
```suggestion
            # Compute the tuning parameter as explained in the notes.
            # It ensures that pt_src[0] is mapped onto pt_dst[0] after transformation.
```

src/daria/corrections/shape/curvature.py
```diff
@@ -387,18 +388,46 @@ def compute_stretch(self, img: Optional[np.ndarray] = None, **kwargs):
...
+        # (within 10% of total pixels), and make warning
+        elif abs(pt_src[1] - stretch_center[1]) < round(0.1 * Ny):
+            vertical_stretch = 0
+            warn(
+                "point_source chosen too close to stretch center for correction"
+                " in vertical direction. Please choose a value towards the corners."
+                " Vertical stretch is set to zero now."
+            )
+        else:
+            # Compute the tuning parameter as explained in the notes
```
jwboth, line 427:
```suggestion
            # Compute the tuning parameter as explained in the notes.
            # It ensures that pt_src[1] is mapped onto pt_dst[1] after transformation. 
```

jwboth (approved):

src/daria/corrections/shape/curvature.py
EStorvik: I made a suggestion. Perhaps you could take a look before we merge. I chose to split it into three parts so that the warning tells the user which "mistake" is being made. Moreover, I adjusted it to 5% of the image's pixels as 10% seemed a bit much. What do you think?

src/daria/corrections/shape/curvature.py
```diff
@@ -398,15 +398,41 @@ def compute_stretch(self, img: Optional[np.ndarray] = None, **kwargs):
...
-        elif abs(pt_src[0] - stretch_center[0]) < round(0.1 * Nx):
+        elif abs(pt_src[0] - stretch_center[0]) < round(0.05 * Nx):
             horizontal_stretch = 0
             warn(
                 "point_source chosen too close to stretch center for correction"
-                " in horizontal direction. Please choose a value towards the corners."
-                " Horizontal stretch is set to zero now."
+                " in horizontal direction (within 5"
+                "%"
+                " of pixels in horizontal "
```
jwboth, line 407: Will this be displayed in three lines?

jwboth (approved):
Looks good. I like using only 5%. Before I used marks on the color checker with known coordinates to define the stretch and it worked. I am not sure whether the color palette falls into 5%, but let's not think about this now. If it becomes relevant, one could add the 5% as a tuning paremeter for the advanced users.

All in all, it is pretty extensive (50 lines for some simple warnings), in paricular because 50% of the warning is used three times. But let's not overthink this now.

thanks for the effort!

src/daria/corrections/shape/curvature.py
EStorvik: I will test it, but I think this is the/a way to display it without having several lines. flake8 is prohibiting me from making just one superlong string..

jwboth: Cool. Whenever you are satisfied, feel free to merge.

EStorvik: It prints as expected. I just removed on whitespace to make it look nicer. Merging now.

PR #86, opened 2022-10-22
Split I/O of the cleaning filter in the concentration analysis

jwboth:
So far `daria.ConcentrationAnalysis` provides I/O for scaling parameters and cleaning filters. However, the combination does not always make much sense (in particular when there is no scaling). Therefore, a split of the two is suggested, and a separate I/O routine for the cleaning filter is provided. The old routines are kept for now (due to use in other projects) but ammended by a deprecation warning. They should be removed at some point, in coordination with updating projects using DarIA.

In addition, small fixes are included (doc, verbosity, and a small non-critical bug).

### DIFF
--- a/src/daria/analysis/concentrationanalysis.py
+++ b/src/daria/analysis/concentrationanalysis.py
@@ -7,6 +7,7 @@ import copy
 import json
 from pathlib import Path
 from typing import Optional, Union
+from warnings import warn
 
 import cv2
 import matplotlib.pyplot as plt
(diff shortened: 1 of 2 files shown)


src/daria/analysis/concentrationanalysis.py
```diff
@@ -470,6 +504,10 @@ def __init__(
     def update_mask(self, mask: np.ndarray) -> None:
         """
         Update the mask to be considered in the analysis.
+
+        Args:
+            mask (np.ndarray): boolean mask, detecting which pixels will
+                be considered, al other will be ignored in the analysis.
```
EStorvik, line 510:
```suggestion
                be considered, all other will be ignored in the analysis.
```

src/daria/analysis/concentrationanalysis.py
```diff
@@ -642,7 +681,7 @@ def _prior(self, signal: np.ndarray) -> np.ndarray:
             plt.imshow(mask)
 
         # Finaly cleaning - deactive signal outside mask
```
EStorvik, line 683:
```suggestion
        # Finaly cleaning - inactive signal outside mask
```

EStorvik (approved):
I do not have any experience working with this, but it sounds (and looks) very reasonable

PR #87, opened 2022-10-22
Reading images and converting to RGB 

jwboth:
So far, when no `color_correction` has been defined in `daria.Image`, no converison to RGB has been applied. When reading already converted images, one does not necessarily want to define and apply empty correction (which still would take time to execute). Hence, the transformation to RGB is applied always.

In `DriftCorrection`, so far `PIL` has been used to read images. In view of the changes in #79 the baseline image in `DriftCorrection` however should also be read in the same format as in `Image`. Thus, also `cv2.imread` is used here, with the same options to read int16 images.

### DIFF
--- a/src/daria/corrections/shape/drift.py
+++ b/src/daria/corrections/shape/drift.py
@@ -9,8 +9,8 @@ import json
 from pathlib import Path
 from typing import Optional, Union
 
+import cv2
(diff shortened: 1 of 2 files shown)


EStorvik (approved):
Good call!

PR #88, opened 2022-10-23
Add first managers

jwboth:
In order to simplify the setup of analysis objects for dedicated studies, managing classes are intended for such use. This PR adds two first objects targeting this purpose. At the moment, a general base class is added, and a class towards tracer analysis. In near future, also an additional object tailored to CO2 analyses will be added.

Important: When setting up all my run scripts to analyze the benchmark data, I have not used the recent keyword choices introduced in the correction objects of daria. Instead, I have been using nested config files, to allow for some structure. As currently implemented, the managers explicitly expect such a nested structure. Thus, a unified choice will have to be made, what style daria is expecting, and all script will have to be adapted. At the moment, this is not the purpose of this PR, so I am keeping the style I have been using.

### DIFF
--- a/src/daria/manager/analysisbase.py
+++ b/src/daria/manager/analysisbase.py
@@ -57,7 +57,7 @@ class AnalysisBase:
             config=self.config["geometry"]
         )
 
-        # Define baseline image as corrected daria Image - if exists use cached image
+        # Define baseline image as corrected daria Image
         self.base = self._read(reference_base)
 
     # ! ----- I/O
--- a/src/daria/manager/analysisbase.py
+++ b/src/daria/manager/analysisbase.py
@@ -1,0 +1,95 @@
+"""
+Module providing structures for common analyses. In practice, such may
+have to be tailored to the specific scenario. Yet, they already provide
+many of the most relevant functionalities. If not applicable, they also
+provide the approach for how to set up tailored analysis classes.
+"""
+
+import json
+from pathlib import Path
+from typing import Union
+
+import daria
+
+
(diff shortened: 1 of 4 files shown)


src/daria/manager/analysisbase.py
```diff
@@ -0,0 +1,95 @@
...
+        # Define correction objects
+        self.drift_correction = daria.DriftCorrection(
+            base=reference_base, roi=self.config["drift"]["roi"]
+        )
+        self.color_correction = daria.ColorCorrection(roi=self.config["color"]["roi"])
+        self.curvature_correction = daria.CurvatureCorrection(
+            config=self.config["geometry"]
+        )
+
+        # Define baseline image as corrected daria Image - if exists use cached image
```
jwboth, line 60:
```suggestion
        # Define baseline image as corrected daria Image
```

EStorvik (approved):
Looks very good! Would looking forward to seeing the full analysis scripts as well!

With regards to the nested config files, I think that is a good idea. I halfway went there in the previous PR as well, but it might just be as easy to "manually" nest them..

jwboth:
> Looks very good! Would looking forward to seeing the full analysis scripts as well!
> 
> With regards to the nested config files, I think that is a good idea. I halfway went there in the previous PR as well, but it might just be as easy to "manually" nest them..

I think there is no obvious way how to do it. Therefore, I am suggesting, we have a discussion altogether when a full example is uploaded somewhere (most likely not here before sorting things out). And then we define a standard.

PR #89, opened 2022-10-23
Add functionality applying the deformation/compaction map to any image

jwboth:
The `CompactionAnalysis` operates on two images, a source and destination image. It can determine the deformation from one to the other and also stores the deformation map. It can be evaluated. This PR adds the functionality to apply this deformation map to any input, which can be also different from the source and destination images. 

Example use case: A use case is/can be the deformation of the segmentation of an baseline geometry of a FluidFlower from one to another scenario. When compaction occurs or a different config file is used for `CurvatureCorrection`, also the segmentation is changing. By applying the computed deformation map, a previously computed segmentation (which may be slightly hardcoded), can be transferred without enterring further tuning etc.

### DIFF
--- a/src/daria/analysis/compactionanalysis.py
+++ b/src/daria/analysis/compactionanalysis.py
@@ -108,3 +108,19 @@ class CompactionAnalysis:
             deformation = base.coordinatesystem.pixelToCoordinateVector(deformation)
 
         return deformation
+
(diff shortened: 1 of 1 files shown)


EStorvik (approved):
I had not thought about the application, but it makes sense, and anyways the function is good to have I think. Very nice!

PR #91, opened 2022-10-26
Improve analysis managers

jwboth:
A CO2 analysis manager has been added, making a standard FluidFlower analysis possible. General bits of the the tracer analysis manager are required for that. Therefore, the tracer analysis has been split, introducing a new base class for general concentration based analyses. The tracer analysis is now less general and instead specialized for tracer analysis.


EStorvik (approved):
I think it looks good, and it seems like a nice way to do things!

PR #92, opened 2022-10-27
Compute contour length of segmentations

jwboth:
Add capability which computes the contour length of a boolean or integer-values array (restricted to prescribed values) and restricted to a ROI. This allows to perform e.g. a fingering analysis for CO2 storage experiments.

As part of this PR (as it is also used in the main change): `extractROI` has been updated, using an alternative approach (more compatible to other places in daria, I believe) to define a ROI. By providing a list/array of coordinates, essentially a bounding box is defined. This can be reverted if not agreed upon.

### DIFF
--- a/examples/notebooks/brief_tutorial.ipynb
+++ b/examples/notebooks/brief_tutorial.ipynb
@@ -76,7 +76,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "If one would rather like to extract a ROI, defines as box with corners [10, 20] and [50, 70] (in physical physical coordinates), this is possible by the use of the command below:"
+    "If one would rather like to extract a ROI, defined as a box with corners [10, 20] and [50, 70] (in physical physical coordinates), this is possible by the use of the command below:"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


examples/notebooks/brief_tutorial.ipynb
```diff
@@ -76,7 +76,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "If one would rather like to extract a ROI thought physical coordinates, in the coordinates 10 to 50 in the x-direction and 20 to 70 in the y-direction this is possible by the use of the command below:"
+    "If one would rather like to extract a ROI, defines as box with corners [10, 20] and [50, 70] (in physical physical coordinates), this is possible by the use of the command below:"
```
EStorvik, line 79:
```suggestion
    "If one would rather like to extract a ROI, defined as a box with corners [10, 20] and [50, 70] (in physical physical coordinates), this is possible by the use of the command below:"
```

EStorvik (approved):
Very cool! Did not know of these skimage functions. Could probably also have been used in the "dragon"-creation.

jwboth:
> Very cool! Did not know of these skimage functions. Could probably also have been used in the "dragon"-creation.

I think indeed that there would have been tools for assisting in finding differences between masks. Whenever I come across them (and after understanding the details of your PR), I may suggest replacing if there is a significant reduction of code.

PR #93, opened 2022-10-27
Update to segmentation comparison

EStorvik:
Created a class structure for the segmentation comparison to help in visualizing the compared images. It should not in any way affect other parts of DarIA, so we could also discuss if it actually should not be a part of DarIA at all. Anyways, I'm creating this PR, and we can have a small discussion. No need to look carefully into the details (although they should not be too complicated anywhere).


src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+    Routines for comparing segmentations and creating visualizations of the comparison.
+
+    Attributes:
+        number_of_segmented_images (int): Number of segmented images that one compares
+        segmentation_names (list[str]): list of names for each of the segmentations.
+            Will affect legends in plots.
+        components (list): list of values the different (active) components in the
+            segmentations. As of now, up to two are allowed for and the default values
+            are 1 and 2.
+        non_active_component (int): value for the nonactive component in the
```
jwboth, line 28: why is this needed, when `components` needs to be provided? Are not all other components inactive?

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+        self.component_names: list = kwargs.pop(
+            "component_names", ["Component 0", "Component 1"]
+        )
+
+        # Define gray colors
+        self.gray_colors: np.ndarray = kwargs.pop(
+            "gray_colors",
+            np.array(
+                [[180, 180, 180], [220, 220, 220], [200, 200, 200]], dtype=np.uint8
+            ),
```
jwboth, line 86: If really only gray colors are intended, wouldn't it be better to only ask for monochromatic values instead of trichromatic?

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                f"{i+2} segmentations overlaps in {self.component_names[0]}"
+            ] = self.gray_base[0] * (i + 1)
+            self.color_dictionary[
+                f"{i+2} segmentations overlaps in {self.component_names[1]}"
+            ] = self.gray_base[1] * (i + 1)
+            self.color_dictionary[
+                f"Mixture {self.component_names[0]} and "
+                f"{self.component_names[1]} for {i+2} segmentations"
+            ] = self.gray_base[2] * (i + 1)
+
```
jwboth, line 141: This only allows for two consecutive images to overlap. When considering however three or more images, there is also other combinations than 0 and 1, as well as 1 and 2. A simple nested for loop structure should solve the problem.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+        """
+        Comparison of segmentations.
+
+        Args:
+            segmentations (asterisk argument): Allows to provide an arbitraty
+                number of segmented numpy arrays or da.Images of integers to be compared
+            Optional keyword arguments (kwargs):
+                plot_result (bool): plots the result with matplotlib if True,
+                    default is False.
+                roi (tuple): roi where the segmentations should be compared,
```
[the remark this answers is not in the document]
jwboth: Also mention that the ROI should be provdide using matrix indexing.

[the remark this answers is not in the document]
jwboth: and addresses pixel coordinates (not metric units).

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                dtype=np.uint8,
+            )
+
+        else:
+            if isinstance(segmentations[0], np.ndarray):
+                nx = min([seg.shape[0] for seg in segmentations])
+                ny = min([seg.shape[1] for seg in segmentations])
+            elif isinstance(segmentations[0], da.Image):
+                nx = min([seg.img.shape[0] for seg in segmentations])
+                ny = min([seg.img.shape[1] for seg in segmentations])
```
jwboth, line 179: This is confusing. nx and ny should be reversed. I woudl rather use `rows` and `cols` to reduce confusion.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+            roi = (slice(0, nx), slice(0, ny))
+            return_image: np.ndarray = np.zeros((nx, ny) + (3,), dtype=np.uint8)
+
+        # Enter gray everywhere there are ovelaps of different segmentations
+
+        if isinstance(segmentations[0], np.ndarray):
```
jwboth, line 185: The `if` and `else` statements are exactly the same up to the way of addressing the arrays. Making use of call by reference, I guess, it would be cleaner to first set up references to the corresponding objects depending on the data type and then have a shorter (and single) analysis.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                    return_image[
+                        np.logical_or(
+                            np.logical_and(
+                                segmentations[k][roi] == self.components[0],
+                                segmentations[i][roi] == self.components[1],
+                            ),
+                            np.logical_and(
+                                segmentations[k][roi] == self.components[1],
+                                segmentations[i][roi] == self.components[0],
+                            ),
```
jwboth, line 208: To prepare for multiple component, it would be simpler to simple check whether segmentations[k][roi] values lie in self.components and segmentations[k][roi] is not equal segmentation[i][roi]. This shoudl be 2-3 lines and generalize.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                        )
+                    ] += self.gray_base[2]
+
+            # Determine locations (and make modifications to return image) of unique components
+            for c_num, c in enumerate(self.components):
+                for k in range(self.number_of_segmented_images):
+                    only_tmp: np.ndarray = segmentations[k][roi] == c
```
jwboth, line 215: I would try to use a more descriptive name and aim at not using anything with `tmp`.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                            np.logical_and(
+                                segmentations[k][roi] == self.components[0],
+                                segmentations[i][roi] == self.components[1],
+                            ),
+                            np.logical_and(
+                                segmentations[k][roi] == self.components[1],
+                                segmentations[i][roi] == self.components[0],
+                            ),
+                        )
+                    ] += self.gray_base[2]
```
jwboth, line 210: To be consistent with the last assignment
```suggestion
                    ] = self.gray_base[2]
```

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                for i in range(k + 1, self.number_of_segmented_images):
+
+                    # Overlap of components
+                    for c_num, c in enumerate(self.components):
+                        return_image[
+                            np.logical_and(
+                                segmentations[k][roi] == c,
+                                segmentations[i][roi] == c,
+                            )
+                        ] += self.gray_base[c_num]
```
jwboth, line 196:
```suggestion
                        ] = self.gray_base[c_num]
```
To be consistent with the last assignment.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+                    for j in filter(
+                        lambda j: j != k, range(self.number_of_segmented_images)
+                    ):
+                        only_tmp = np.logical_and(
+                            only_tmp,
+                            segmentations.img[j][roi] == self.non_active_component,
+                        )
 
-            return_image[only_tmp] = colors[k, c_num]
+                    return_image[only_tmp] = self.colors[k, c_num]
```
jwboth, line 264: I would try to remove this. This should be fixed by one line of code.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+        # Get unique colors and sort them
+        unique_colors = self._get_unique_colors(comparison_image)
+        unique_colors = self._sort_colors(unique_colors)
+
+        # Process the comparison image
+        processed_comparison_image = self._post_process_image(
+            comparison_image,
+            unique_colors=unique_colors,
+            opacity=opacity,
+            countor_thickness=15,
```
jwboth, line 460:
```suggestion
            contour_thickness=15,
```

jwboth (approved):
I think it looks OK. I think one could improve the readibility by some more comments, and simplify the code here and there. But tit seems OK. Did you try the code with more than two images?

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+import matplotlib.pyplot as plt
 import numpy as np
+import skimage
 from matplotlib.cm import get_cmap
 
+import daria as da
+
 
-def compare_segmentations(*segmentations, **kwargs) -> np.ndarray:
+class CompareSegmentations:
```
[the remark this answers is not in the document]
EStorvik: Haha, I found it extremely much simpler in the end. Of course you were right all along! I totally agree on the name change. I don't know why I haven't changed it myself yet... at least three times I mistakenly called it using that name when I played around with it..

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,471 @@
...
+    def _post_process_image(
+        self,
+        image: np.ndarray,
+        unique_colors: Optional[np.ndarray] = None,
+        opacity: float = 1.0,
+        countor_thickness: int = 10,
+    ) -> np.ndarray:
+        """
+        Routine for post processing the image, removing background,
+        drawing countors and applying opacity.
```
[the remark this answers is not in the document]
EStorvik: Hahaha, clearly I have no clue how to spell contour...

src/daria/analysis/segmentationcomparison.py
EStorvik: Sounds very correct, ill remove it!

src/daria/analysis/segmentationcomparison.py
EStorvik: I thought about it, but decided that probably it could be a good idea to have the option to have other colors than gray in the future when more segmentations are compared. Moreover, I am not sure how easy it is to combine monochromatic and rgb colors in the same image?

src/daria/analysis/segmentationcomparison.py
EStorvik: I don't understand what you mean? This is just the setup of the colors, it doesn't actually check which images overlap. Nevertheless, if we really want to distinguish the different gray colors for more than two images I do not think that the code can be kept like this, and we probably need to just choose specific color values for the cases we want to highlight.

src/daria/analysis/segmentationcomparison.py
EStorvik: Good call!

src/daria/analysis/segmentationcomparison.py
EStorvik: I made a suggestion now. It felt a bit more scary as it will return true if segmentations[i][roi] is a non active component where segmentations[k][roi] is in self.components, but that case is overwritten when checking for unique apperances.

src/daria/analysis/segmentationcomparison.py
EStorvik: same?

EStorvik:
I made several modifications according to your suggestions, and one very minor one in daria.Image. I was a bit afraid to put it here, but I just moved the toRGB() line, and thought it was ok to not make an entirely new pr for it. As it was implemented it didn't allow for monochromatic images to initilize the daria image, now it should be fine. If you disagree, ill remove the change from here, and make a separate PR. let me know.

jwboth (approved):
Thanks for simplifying the code and additing comments. I think it looks much better now!

EStorvik:
I do too, thanks a lot!

src/daria/analysis/segmentationcomparison.py
EStorvik: I am not sure what you mean?

EStorvik:
Do you want to check over the unresolved conversations before I merge?

src/daria/analysis/segmentationcomparison.py
jwboth: I did not get why you use summation here? after all, it seemed you treat disjoint cases anyhow. Overwriting seemed most natural. If I got something wrong, simply ignore my comment.

src/daria/analysis/segmentationcomparison.py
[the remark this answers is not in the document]
jwboth: Looked OK, when I read it.

src/daria/analysis/segmentationcomparison.py
EStorvik: My "secret" plan here, is that with the summation I can check how many are overlapping, and not just whether some are overlapping or not. I am almost sure that it will be too much information in the end (with 5 compared images), but I will keep it for now, and probably revert to the equality at a later point.

EStorvik: I tested with 4 images now, and it became a true mess. I will make the equality instead.

src/daria/analysis/segmentationcomparison.py
EStorvik: This was indeed a mess, and I fixed it now.

jwboth: ah OK, I misunderstood the code I think. Sorry. I guess a comment could have helped, but it is fine.

src/daria/analysis/segmentationcomparison.py
jwboth: Ah OK. In principle I am fine with these type of hacks. But then, good documentation is needed - for the future hacks.

src/daria/analysis/segmentationcomparison.py
```diff
@@ -1,113 +1,442 @@
...
+        # Determine whether segmentations are arrays of daria images.
+        # They should all be the same.
+        if all([isinstance(seg, np.ndarray) for seg in segmentations]):
+            segmentation_arrays: tuple[np.ndarray, ...] = segmentations
+        elif all([isinstance(seg, da.Image) for seg in segmentations]):
+            segmentation_arrays = tuple([seg.img for seg in segmentations])
+        else:
+            raise Exception(
+                "Segmentation types are not allowed. They should"
+                "all be the same, and either arrays, og daria images."
```
jwboth, line 197:
```suggestion
                "all be the same, and either arrays, or daria images."
```

jwboth:
> Do you want to check over the unresolved conversations before I merge?

Looks all good to me. Feel free to resolve all conversations and merge.

PR #94, opened 2022-10-27
Timestamp can be fed to image in initialzation

EStorvik:
Included the possibility for Image to read timestamps. There is no explicit check that they are in datetime.datetime format. Do you think that should be added? I am honestly not to familiar with this.

### DIFF
--- a/src/daria/image/image.py
+++ b/src/daria/image/image.py
@@ -43,7 +43,7 @@ class Image:
             origo (np.ndarray): physical coordinates of the lower left corner, i.e.,
                 of the (img.shape[0],0) pixel
             color_space (str): Color space (RGB/BGR/RED/GREEN/BLUE/GRAY)
-            timestamp (datetime.datetime): timestamp of the image in datetime format.
+            timestamp (datetime.datetime): timestamp of the image in (default) datetime format.
         shape (np.ndarray): num_pixels, as well number of color channels (typically 3 for RGB)
         dx (float): pixel size in x-direction
         dy (float): pixel size in y-direction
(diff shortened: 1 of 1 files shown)


src/daria/image/image.py
```diff
@@ -43,6 +43,7 @@ class Image:
             origo (np.ndarray): physical coordinates of the lower left corner, i.e.,
                 of the (img.shape[0],0) pixel
             color_space (str): Color space (RGB/BGR/RED/GREEN/BLUE/GRAY)
+            timestamp (datetime.datetime): timestamp of the image in datetime format.
```
jwboth, line 46:
```suggestion
            timestamp (datetime.datetime): timestamp of the image in (default) datetime format.
```

src/daria/image/image.py
```diff
@@ -139,7 +140,7 @@ def __init__(
             self.metadata["color_space"] = "BGR"
             self.metadata["original_dtype"] = self.img.dtype
 
-            if get_timestamp:
+            if self.metadata["timestamp"] is None:
                 pil_img = PIL_Image.open(Path(img))
 
                 # Read exif metadata
```
jwboth, line 146:
```suggestion
                # Read timestamp from exif metadata if existent
```

jwboth (approved):

jwboth:
I think it looks good for now. Let's see if adaptions are needed when we start using this. My recommendation for these type of updates is to simulateneously use the functionality in a use case to see whether the final solution is working out. This leads to less iterations. This is so short, that I would hope that everything works out.

EStorvik:
I agree, but I didn't bother writing the code for transcribing the image names to datetime as Benyamine told us that he already has a plan for it. I think/hope that this will not affect any current code, and we will quickly see if everything works out when Benyamine tries it out. You can merge if you think it is ok!

PR #95, opened 2022-10-27
Add possibility to address negatives of boolean masks

jwboth:
It turned out to be useful to be able to compute the contour line of the logical negative of a boolean image. Therefore this possibility has been added to the contour analysis.

### DIFF
--- a/src/daria/analysis/contouranalysis.py
+++ b/src/daria/analysis/contouranalysis.py
@@ -48,7 +48,7 @@ def contour_length(
         if values_of_interest is None:
             mask: np.ndarray = img_roi.img
         else:
-            mask = np.zeros(img_roi.img.shape[:2], dtype=bool)
+            mask = np.zeros_like(img_roi.img, dtype=bool)
             for value in values_of_interest:
                 mask[img_roi.img == value] = True
     elif img_roi.img.dtype in [np.uint8, np.int32, np.int64]:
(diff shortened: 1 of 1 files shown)


src/daria/analysis/contouranalysis.py
```diff
@@ -45,7 +45,12 @@ def contour_length(
 
     # Extract boolean mask covering pixels of interest.
     if img_roi.img.dtype == bool:
-        mask: np.ndarray = img_roi.img
+        if values_of_interest is None:
+            mask: np.ndarray = img_roi.img
+        else:
+            mask = np.zeros(img_roi.img.shape[:2], dtype=bool)
```
EStorvik, line 51: why are you taking the .shape[:2] here, and not above (in the if-statement)? If they are supposed to yield the same sized array, why not use zeros_like?

EStorvik (approved):

src/daria/analysis/contouranalysis.py
jwboth, line 51:
```suggestion
            mask = np.zeros_like(img_roi.img, dtype=bool)
```

jwboth: It has become a habit to not use zeros_like. Failed to often assuming a wrong shape when working with images, I guess. Here it is safe though. I changed it, thanks.

PR #96, opened 2022-10-27
Add possibility to determine the perimeter of a bounding box

jwboth:
In the contour analysis it is useful to determine the contour of a box. As this can be easily standardied and at the same takes some lines, it is suggested that this functionality is added to daria directly.

### DIFF
--- a/src/daria/utils/box.py
+++ b/src/daria/utils/box.py
@@ -77,14 +77,16 @@ def perimeter(box: Union[tuple, np.ndarray]) -> Union[int, float]:
     Returns:
         float or int (depending on input): perimeter
     """
-    if isinstance(box, np.ndarray):
-        min_x = np.min(box[:, 0])
-        max_x = np.max(box[:, 0])
-        min_y = np.min(box[:, 1])
-        max_y = np.max(box[:, 1])
-        perimeter = 2 * (max_x - min_x) + 2 * (max_y - min_y)
+    # Convert to array
+    box = box if isinstance(box, np.ndarray) else bounding_box_inverse(box)
 
-    elif isinstance(box, tuple):
-        raise NotImplementedError("This is currently not supported.")
+    # Extract min and max for each dimension
+    min_x = np.min(box[:, 0])
+    max_x = np.max(box[:, 0])
+    min_y = np.min(box[:, 1])
+    max_y = np.max(box[:, 1])
+
+    # Determine perimeter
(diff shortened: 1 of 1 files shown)


src/daria/utils/box.py
```diff
@@ -62,3 +62,29 @@ def bounding_box_inverse(bounding_box: tuple) -> np.ndarray:
...
+    """
+    if isinstance(box, np.ndarray):
+        min_x = np.min(box[:, 0])
+        max_x = np.max(box[:, 0])
+        min_y = np.min(box[:, 1])
+        max_y = np.max(box[:, 1])
+        perimeter = 2 * (max_x - min_x) + 2 * (max_y - min_y)
+
+    elif isinstance(box, tuple):
+        raise NotImplementedError("This is currently not supported.")
```
EStorvik, line 88:
```suggestion
       box_array =  bounding_box_inverse(box)
       min_x = np.min(box_array[:, 0])
       max_x = np.max(box_array[:, 0])
       min_y = np.min(box_array[:, 1])
       max_y = np.max(box_array[:, 1])
       perimeter = 2 * (max_x - min_x) + 2 * (max_y - min_y)
       
```

EStorvik: I just wrote this in github now, not 100% sure if it is exactly what you want. Just disregard it if that is the case.

EStorvik (approved):
Seems like a very reasonable functionality to have!

src/daria/utils/box.py
```diff
@@ -62,3 +62,29 @@ def bounding_box_inverse(bounding_box: tuple) -> np.ndarray:
...
+        float or int (depending on input): perimeter
+    """
+    if isinstance(box, np.ndarray):
+        min_x = np.min(box[:, 0])
+        max_x = np.max(box[:, 0])
+        min_y = np.min(box[:, 1])
+        max_y = np.max(box[:, 1])
+        perimeter = 2 * (max_x - min_x) + 2 * (max_y - min_y)
+
+    elif isinstance(box, tuple):
```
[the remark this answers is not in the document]
jwboth: Good idea.

src/daria/utils/box.py
jwboth: I took the libert to disregard this suggestion due to code duplication. Instead, unified conversion is provided at the start of the routine.

PR #97, opened 2022-10-27
Provide setter for timestamp of Image.

jwboth:
Optimally, metadata should be created at the initiation of an `Image`. However, some of the metadata is not crucial and may even be set to default values as `None`. This also holds for `timestamp`. For such properties it is not a big deal to add a setter method. This PR adds a setter method for `timestamp`. It allows to ammend the metadata of an `Image` with a corresponding time, also after the initialization. This turns out to be a very convenient way of reading images with times, e.g. hardcoded in the file names.

### DIFF
--- a/src/daria/image/image.py
+++ b/src/daria/image/image.py
@@ -193,6 +193,8 @@ class Image:
         # Establish a coordinate system based on the metadata
         self.coordinatesystem: da.CoordinateSystem = da.CoordinateSystem(self)
 
+    # ! ---- Fast-access getter functions for metadata
+
     @property
     def origo(self) -> list:
         return self.metadata["origo"]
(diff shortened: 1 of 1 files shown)


src/daria/image/image.py
```diff
@@ -217,6 +219,12 @@ def timestamp(self) -> datetime.datetime:
     def original_dtype(self) -> np.dtype:
         return self.metadata["original_dtype"]
 
+    # ! ---- Corresponding setter functions for metadata
+
+    @timestamp.setter
```
EStorvik, line 224: never seen this @ commant before. does it work s one would imagine, i.e.,:
#initialize image 
img = da.Image(...)
#set time (where time is in the correct format of course)
img.timestamp = time

EStorvik (approved):
This sounds like a very reasonable functionality to have! Great! I'm sure it can also be useful for other attributes as well in the future.

src/daria/image/image.py
jwboth: This is exactly what this allows. Without the setter method (provided the extra @ command), it is not allowed to access a defined @property.

PR #98, opened 2022-10-28
Construct required foder structure before writing to file

jwboth:
Instead of making the user requiring to provide exitent folder structure when wanting to write to file, make the directories automatically.


EStorvik (approved):
Looks good!

PR #99, opened 2022-10-28
Modified examples and corrections

EStorvik:
Made minor modifications to correction notebook and correction initiaization such that they overlap to a greater extent with the AnalysisBase assumptions on the config setup.

### DIFF
--- a/examples/images/config.json
+++ b/examples/images/config.json
@@ -1,6 +1,6 @@
 {
-    "color_correction": {
-        "roi_color_correction": [
+    "color": {
+        "roi": [
             [
                 146,
                 196
@@ -19,8 +19,8 @@
             ]
         ]
     },
-    "drift_correction": {
-        "roi_drift_correction": [
+    "drift": {
+        "roi": [
             [
                 57,
                 2
@@ -39,7 +39,7 @@
             ]
         ]
     },
-    "curvature_correction": {
+    "geometry": {
         "init": {
             "horizontal_bulge": 5e-09,
             "horizontal_center_offset": 0,
--- a/examples/images/config.json
+++ b/examples/images/config.json
@@ -39,7 +39,7 @@
             ]
         ]
     },
-    "geometry": {
+    "curvature": {
         "init": {
             "horizontal_bulge": 5e-09,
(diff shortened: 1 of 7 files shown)


examples/images/config.json
```diff
@@ -39,7 +39,7 @@
             ]
         ]
     },
-    "curvature_correction": {
+    "geometry": {
```
jwboth, line 42: I wonder whether it would not be actually better to use consistently `curvature_correction`, as we did not change the name of `CurvatureCorrection`.

src/daria/corrections/shape/curvature.py
```diff
@@ -76,19 +76,11 @@ def __init__(
...
+                    self.config = json.load(openfile)
             else:
                 assert isinstance(config, Path)
                 with open(config, "r") as openfile:
-                    tmp_config = json.load(openfile)
-                if "curvature_correction" in tmp_config:
-                    self.config = tmp_config["curvature_correction"]
-                else:
-                    self.config = tmp_config
+                    self.config = json.load(openfile)
```
jwboth, line 83: This is redundant. If `config` is of type `Path`, then `Path(config)` is still of type `Path`. So the case that`config` is of type `str` or `Path` can be covered by the above case.

jwboth (approved):
Looks good aside of the failing test. I think, maybe we should keep `curvature_correction` or call it simply `curvature` to be consistent with `color` and `drift`. What do you think?

src/daria/corrections/shape/curvature.py
EStorvik: I removed it thanks

examples/images/config.json
EStorvik: I changed it to curvature to be consistent with drift and color. Will you modify in the analysis scripts so that they by default looks for curvature and not geometry? Or should I just go into analysisbase.py and change it?

jwboth: I can do it.

src/daria/corrections/shape/curvature.py
jwboth: Great!

PR #100, opened 2022-10-31
Initialize mypy

EStorvik:
Have been fixing annotations and other small mistakes in src with the aim of introducing mypy to the workflow.

There are currently two problems left:

- src\daria\manager\concentrationanalysisbase.py:51: error: Argument 1 to "find_cleaning_filter" of "ConcentrationAnalysis" has incompatible type "Union[str, Path, List[str], List[Path]]"; expected "List[Image]"
- src\daria\analysis\translationanalysis.py:78: error: Missing positional argument "base" in call to "update_base" of "TranslationAnalysis"

Both are somehow mistakes. 

The first one feeds paths while the function expects initialized images. I am not sure how you want this to be. Should the images be initialized in base, or should the function accept paths and initialize itself?

The second one is a missing argument for updating the base image in translationanalysis. I was a bit unsure what your thoughts are here. Currently there is no base attribute in the object, and I am not sure if you want there to be. If not the "base" should probably be fed as an input to the update_params function.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,9 @@ jobs:
       if: ${{ always() }}
       run: flake8 src
 
-#    - name: mypy
-#      if: ${{ always() }}
-#      run: mypy src
+    - name: mypy
+      if: ${{ always() }}
+      run: mypy src
 
     - name: isort
       if: ${{always()}}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,9 +52,9 @@ jobs:
       if: ${{ always() }}
       run: flake8 src
 
-    - name: mypy
-      if: ${{ always() }}
-      run: mypy src
+    # - name: mypy
+    #   if: ${{ always() }}
+    #   run: mypy src
 
     - name: isort
       if: ${{always()}}
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,47 @@
+[mypy]
+mypy_path = src/daria
+
+
+# It is not at all clear to me what to do with various libraries such as numpy etc.
+# For the moment, ignore some of them (seems to be done as below), and hope my understanding
+# improves in the future. 
+
+# Open CV
+[mypy-cv2]
+ignore_missing_imports = True
+
+# Skimage
+[mypy-skimage]
+ignore_missing_imports = True
+
+# sklearn
+[mypy-sklearn.linear_model]
+ignore_missing_imports = True
+
+
+# Matplotlib
+[mypy-matplotlib.pyplot]
+ignore_missing_imports = True
+[mypy-matplotlib.cm]
+ignore_missing_imports = True
+[mypy-matplotlib]
+ignore_missing_imports = True
+
+# Scipy
+[mypy-scipy]
+ignore_missing_imports = True
+[mypy-scipy.sparse]
+ignore_missing_imports = True
+[mypy-scipy.sparse.linalg]
+ignore_missing_imports = True
+[mypy-scipy.ndimage]
+ignore_missing_imports = True
+[mypy-scipy.interpolate]
+ignore_missing_imports = True
+[mypy-scipy.optimize]
+ignore_missing_imports = True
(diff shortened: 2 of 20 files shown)


jwboth:
Just had a quick look and saw a predominant use of `cast`. I do not know this - we never used this in PorePy for instance. Is this a "dirty fix" for type incompatibility?

EStorvik:
It might be. I am actually not sure. Most places can also be fixed with assert, but I saw someone suggesting the use of cast while googling. There might also be even better solutions, but there were about 150 errors when I started, so I prioritized getting through them all.

jwboth:
Rgarding the first point of yours. I think a modular approach is most flexible. Thus, `ConcentrationAnalysis` should not take care of any corrections. The user has to make sure that the baseline images that are used for setting up the cleaning filter are compatible with the images that are later similarly feeded to `ConcentrationAnalysis`, namely preprocessed. Thus, it actually does not make any sense to provide paths to this object at all, at no stage. I am trying to understand the error message you posted from the concentration analysis manager, and I do not fully understand how the type of baseline (from analysis base) propagated to processed_baseline_images...

jwboth:
> 

As long as `cast` does not do anything dangerous and changes formats etc. maybe this is a first fix. This is the danger of not starting with mypy way from the beginning I guess :)

jwboth:
It is incredible how long it takes to run mypy! I am extremely surprised and wonder what is the reason for that.

jwboth:
The second error you mentioned is fixed by the commit I added. Indeed it was not menat as it was implemented.

EStorvik:
I have no idea what is normal for mypy, but indeed it has been superslow all day... It might be that I included some errors when trying to fix the types... In that case, I'm sorry! 
cast should not do anything from https://docs.python.org/3/library/typing.html:
typing.cast(typ, val)
Cast a value to a type.

This returns the value unchanged. To the type checker this signals that the return value has the designated type, but at runtime we intentionally don’t check anything (we want this to be as fast as possible).

So I guess it is really like a hack, but I found it a bit difficult to manage the Unions when mypy didn't accept the if isinstance checks as sufficient..

jwboth:
> I have no idea what is normal for mypy, but indeed it has been superslow all day... It might be that I included some errors to the ConcentrationAnalysis when trying to fix the types... In that case, I'm sorry! cast should not do anything from https://docs.python.org/3/library/typing.html: typing.cast(typ, val) Cast a value to a type.
> 
> This returns the value unchanged. To the type checker this signals that the return value has the designated type, but at runtime we intentionally don’t check anything (we want this to be as fast as possible).
> 
> So I guess it is really like a hack, but I found it a bit difficult to manage the Unions when mypy didn't accept the if isinstance checks as sufficient..

Sounds like a dirty fix after all. The right way would be to fix the Unions... or get rid of Unions and be not too flexible in the inputs. Assert is often indeed needed as you mentioned.

It is of course not easy to debug if mypy requires super long time to be run - mypy on entire porepy requires less than 2 minutes, and porepy is larger than daria.

I will have to double check whether concentration analysis gets broken by this or not. I should have access to suitable test cases.

EStorvik:
I found this while googling:

https://github.com/python/mypy/issues/3796

We have some dictionaries around.. Seems quite rediculous, but it could be one problem I guess.

jwboth:
After you have run mypy for daria on your system. Do you still anticipate long execution times (mypy caches previous results)? If it is only github that is struggling with executing mypy, maybe we can live with that?

EStorvik:
Everytime I run mypy (even for single files) it takes a long time. Running just on one file takes me about a minute. For repositories,  even longer. I am not sure what I do wrong.

EStorvik:
Small update: I tried deleting the mypy cache, and then it took over 6 minutes to run mypy on my computer. After the cache was remade it took less than a minute. I guess that is fine? Still a bit annoying to wait a minute every time, but I guess that is reasonable fast?

EStorvik:
I'm merging with (oral) approval

jwboth (approved):
Great effort!

PR #101, opened 2022-11-02
Bugfix in segmentationcomparison.

EStorvik:
When I included the possibility to have the roi as an array, an unfortunate "redefinition" of the roi led to malfunction. It is working as intended now now.

### DIFF
--- a/src/daria/analysis/segmentationcomparison.py
+++ b/src/daria/analysis/segmentationcomparison.py
@@ -163,25 +163,28 @@ class SegmentationComparison:
         # Define number of segmentations
         assert self.number_of_segmented_images == len(segmentations)
 
+        # Checks whether roi is provided and if it is as a tuple (of slices)
+        # or an array of corner points
         if "roi" in kwargs:
-            roi = kwargs["roi"]
-            if isinstance(roi, np.ndarray):
-                roi_tmp = da.bounding_box(roi)
-            elif isinstance(roi, tuple):
-                roi_tmp = roi
+            roi_input = kwargs["roi"]
+            if isinstance(roi_input, np.ndarray):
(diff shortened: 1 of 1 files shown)


EStorvik:
At the same time, I managed to get rid of some "cast" usage.

EStorvik:
Sorry, no, this is based on main, and the mypy-PR has NOT been merged. I just thought it would be a good idea when making new updates that mypy runs without complaints on those files.

jwboth:
I see, very good!

src/daria/analysis/segmentationcomparison.py
```diff
@@ -163,25 +163,27 @@ def __call__(
...
-            roi = kwargs["roi"]
-            if isinstance(roi, np.ndarray):
-                roi_tmp = da.bounding_box(roi)
-            elif isinstance(roi, tuple):
-                roi_tmp = roi
+            roi_input = kwargs["roi"]
+            if isinstance(roi_input, np.ndarray):
+                roi: tuple = da.bounding_box(roi_input)
+            elif isinstance(roi_input, tuple):
+                roi = roi_input
```
jwboth, line 173: I usually like to add some error indicating that something went wrong with the data type if someting else is provided. I am not 100% following where the kwargs come from. If they are steered from a json file, including `list` as type may be also relevant, but you know better.

jwboth (approved):

src/daria/analysis/segmentationcomparison.py
EStorvik: Good call. I made both of those changes now. I am not sure if we ever are going to steer it from a json file, but at least we can do it now.

PR #104, opened 2022-11-03
Provide simple tool to screen HSV spectrum of images in different ROIs.

jwboth:
This is a very simple util tool. Not sure whether we want to keep this or extend. That is why I did not add this to daria. So far. However, having the major renaming issue in mind, I am adding it for now.


EStorvik (approved):
I see what you mean, but I think it is ok to have it. No need to advertise for it, but its better for the users to be able to call this function rather than remake it themselves.

PR #105, opened 2022-11-08
Tracer analysis for segmented images

jwboth:
Add dedicated concentration and tracer analysis objects for segmented media. These utilize label data. The files are still under current development and will be changed. In order to facilitate a name change of the repo, the files are however added already now at an intermediate (and working) stage - they can be in part used to analyse the well test for the FluidFlower Benchmark.


EStorvik:
How careful do you want me to be here?

jwboth:
> How careful do you want me to be here?

Not at all. This class is quite divergent from the rest and if things break, it is not crucial in other places at the moment. I should also be the only one working with these classes. There will be updates in the near future. Yet, I thought, let's not wait for these before renaming.

EStorvik (approved):
Looked very superficially through it, and think it looks good!

PR #106, opened 2022-11-08
Store the curvature correcion operation to file

jwboth:
Allow caching of the curvature correction transformation to file. This is a small operation, when running an analysis once. However, on each initialization of an analysis, the transformation is recomputed. By caching, one just has to read from file. It has to be seen most as a comfort for code development than a necessary feature.

### DIFF
--- a/src/daria/corrections/shape/curvature.py
+++ b/src/daria/corrections/shape/curvature.py
@@ -68,9 +68,6 @@ class CurvatureCorrection:
                 in_meters (bool): returns True if width and height are given
                             in terms of meters. Only relevant if image
                             is provided.
-                use_cache (bool): flag controlling whether the tranformation
-                            map should be fetched from cache.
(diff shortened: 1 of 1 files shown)


src/daria/corrections/shape/curvature.py
```diff
@@ -116,6 +119,11 @@ def __init__(
 
         # Initialize cache for precomputed transformed coordinates
         self.cache = {}
+        self.use_cache = self.config.get("use_cache", False)
```
EStorvik, line 122: I guess that you have tested this, but what is happening here? it looks to me like you would need to get use_cache via the config, although you suggest above that it could be fed through the kwargs?

EStorvik (approved):
I think it is a good idea! See the one comment.

src/daria/corrections/shape/curvature.py
jwboth: You are right. Thanks!

PR #107, opened 2022-11-08
Hotfix: Specify interpolation method in resizing of label data.

jwboth:
Label data is of type integer. When simpy using resize, to enlargen the image, interpolation is used by default creating new unwanted values. By fixing the interpolation method to the crude but here correct nearest neghbor routine, no artificial values are created. This in fact loosens the need (completely most likely) to use dilation in the geometry segmentation which is the current bottleneck of the segment routine.


EStorvik (approved):
Looks great!

PR #108, opened 2022-11-08
Rename: DarSIA

EStorvik:
I tried to find every daria and rename.


EStorvik:
Yes, worked fine! I think that this is done in the git action as well, so we would have been told if it did not work here as well.

jwboth:
Great. Then I guess we are ready to merge and can rename the repo as well.

EStorvik:
Perfect! Just knocked on your door to ask you about this. Ill do it now then!

PR #109, opened 2022-11-09
Include isotropic Bregman TVD as option

jwboth:
The `BinaryConcentrationAnalysis` allowed so far merely for the use of Chambolle and anisotropic Bregman split. Now the option for using isotropic Bregman is included as well.


EStorvik (approved):
Looks good!

PR #112, opened 2022-11-11
Hotfix: Curavture correction without config

jwboth:
When initiating CurvatureCorrection with an image and without a config, obviously config is not accessible. The current code breaks as it tries to access self.config, which is not set. A fix has been added in this PR.


EStorvik (commented):
Very good, did not think of this.

jwboth:
> Very good, did not think of this.

For completeness, can you approve @EStorvik

PR #113, opened 2022-11-13
Sign in compaction analysis

jwboth:
In principle an argument had not been propagated correctly for the compaction analysis. Some more small mistakes have been noticed. The example has been updated accordingly and is producing consistent results.


jwboth (commented):
@EStorvik Since this PR is urgent, I am merging now. Feel free to review this. As the changes are however not critical and should all in all fix some issues, I suspect you will be fine with most of it. If you see potential for crucial improvement, shoot.

PR #114, opened 2022-11-13
Simple clean up of geometry segmentation

jwboth:
The main motivation for this PR is the unified use of resize routines in the segmentation routine. In addition, small adjustments have been made, including unique variable names, and changes in the plotting and default keyword arguments.


EStorvik (approved):
Looks very good!

PR #116, opened 2022-11-17
Extension of concentration analysis

jwboth:
Several changes have been made to the concentration analysis toolbox. The most relevant:
- anisotropic resizing allowing to take into account the presence of thin structures (parallel to x or y axes)
- binary concentration analysis which takes into account information of a segmentation of the image and allows for heterogeneous thresholding
- a suite of dynamic thresholding algorithms


EStorvik (approved):
This was a lot to read! It looks good, but I have not been thorough. I can spend longer if you want me to, but it might be more productive if you just merge and we keep on working? What do you think?

PR #117, opened 2022-11-17
Random fixes

jwboth:
Minor changes. The only crucial was a converison of the data type before performing feature detection. Otherwise, an error had been thrown.

### DIFF
--- a/src/darsia/corrections/shape/curvature.py
+++ b/src/darsia/corrections/shape/curvature.py
@@ -122,6 +122,7 @@ class CurvatureCorrection:
                 self.cache_path = Path(
                     self.config.get("cache", "./cache/curvature_transformation.npy")
                 )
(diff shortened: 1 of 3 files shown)


src/darsia/corrections/shape/curvature.py
```diff
@@ -122,6 +122,7 @@ def __init__(
                 self.cache_path = Path(
                     self.config.get("cache", "./cache/curvature_transformation.npy")
                 )
+                self.cache_path.parents[0].mkdir(parents=True, exist_ok=True)
```
EStorvik, line 125: Good that you added this! I manually made this directory when I ran the code earlier..

EStorvik (approved):
I see you just fixed the if/else things, I was ging to comment about it! It looks good!

PR #119, opened 2022-11-22
Added possibility to compute fractions in comparison images

EStorvik:
Martin requested this functionality. Returns a dictionary where each instance in the comparison is related to the fraction between that instance and all instances.


EStorvik:
I additionally now fixed a contour bug that Benyamine found, and added the possibility to customize the legend.

PR #120, opened 2022-11-23
Scales the base_image in plot overlay

EStorvik:
Now scales the base image for plot overlay (request from Benyamine). Could probably have been done externally as well, but I see no real reason for why it is not included here.


EStorvik:
Im merging this now so that Benyamine can create more dragons with ease.

PR #123, opened 2022-11-24
Provide default options for co2 analysis

jwboth:
Allow to initiate CO2Analysis bjects with incomplete config files. Useful e.g., when only wanting to run parts of the base class but not the actual co2 analysis.


EStorvik (approved):
If you do not provide a "cleaning_filter" and the "cache/cleaning_filter_co2.npy" doesn't exist, what happens? (I am sure that the answer is inside _setup_concentration_analysis, and if you are satisfied with the result in that case it is good for me!)

jwboth:
Yes it will be constructed if it does not already exist.

PR #124, opened 2022-11-24
Add possibility to use TVD instead of median for segmentation.

jwboth:
In addition, apply contrast enhancement, and allow in principle to aplpy segmentation only on parts of the domain.

### DIFF
--- a/src/darsia/utils/segmentation.py
+++ b/src/darsia/utils/segmentation.py
@@ -2,7 +2,7 @@
 Module containing utils for segmentation of layered media.
 """
 
-from typing import Union
+from typing import Optional, Union
 from warnings import warn
 
 import cv2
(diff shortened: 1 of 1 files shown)


src/darsia/utils/segmentation.py
```diff
@@ -33,9 +34,12 @@ def segment(
         edges_method (str): "gradient_based" or "scharr", deciding which algorithm
             is used for determining edges; the former uses gradient filtering while
             the latter uses the Scharr algorithm.
+        mask (np.ndarray, optional): binary array, only where true, segmentation is
+            performed
         verbosity (bool): flag controlling whether relevant quantities are plotted
             which is useful in the tuning of the parameters; the default is False.
         keyword arguments (optional): tuning parameters for the watershed algorithm
+            "method" (str): 'median' or 'tvd'
```
EStorvik, line 42: I am not sure whether it matters, but you could consider adding that tvd is anisotropic.

EStorvik (approved):
Looks good!

EStorvik:
I assume that you will make the checks pass before you merge!

PR #125, opened 2022-11-25
Default setup of folder and standard baseline for colorcorrection

jwboth:
Circumvent error message when path not existent.

The color correction uses now the reference baseline image as default when setting up a custom color checker. This allows for cleaner config files.


EStorvik (approved):
Looks good!

PR #127, opened 2022-11-25
Partial clean up at boundaries under segmentation

jwboth:
Add possibility to steer which boundaries should be cleaned (by constant extension) in the geometry segmentation. Artefact often only appear partially and unwanted overwriting can occur. This can now be controlled with the new option.


EStorvik (approved):
Looks good to me!

PR #128, opened 2022-11-25
Hotfix/Bugfix for segmentation comparison

EStorvik:
A bugfix for contours in the segmentation comparison


jwboth (approved):
Sorry. I thought I already reviewed this... looks good! I assume it works in your examples. I don't exactly understand what went wrong before the fix. Something with monochromatic vs trichromatic colors?

EStorvik:
The "Logic" was wrong and it ended up not finding any contours. I made some adjustments A time back to fix another bug (can't remember the problem then), and accidentally messed up with the contours. I tested it properly on Friday, so I'm quite sure that it works now!

PR #136, opened 2022-12-12
Renamed all instances of origo to origin

EStorvik:
For some reason I assumed that origo was an English word, but it looks like it is not. I changed (hopefully) all apperances of the word in DarSIA. 

I am not sure whether it is used in any external library (fluidflower/porotwin) so perhaps we need to take care?


jwboth (approved):
Looks good to me. I think in the fluidflower context it has not been explicitly addressed anywhere yet. The lower left corner always has been the (0,0) coordinate. So this change is safe.

PR #139, opened 2022-12-17
Overhaul compaction analysis

jwboth:
Not finished work yet, but included for reference: New compaction analysis.


jwboth:
This PR replaces the previous CompactionAnalysis with a general ImageRegistration and integrates the deformation detection as part of the standard deformation correction in the initialization of darsia.Image.

PR #140, opened 2022-12-17
Contour analysis and major overhaul of concentration analysis

jwboth:
Not finished yet, but included for reference: Functionality to count, and track finger tips in spcae-time. etc.


jwboth:
@EStorvik I am expecting you are not intending to take a look. Note though a started restructuring of the folder mathematics in more meaningful folders. The rest touches upon concentration analysis. It is now build in a much more modular way. All specific components have been moved to external classes, as conversion models, regularization, transformation classes, or calibration models. These can be combined in various ways allowing to quickly define new data conversions etc. Examples are currently only the benchmark analysis - the old co2 and tracer test is running, but is not really showcasing the flexibility.

jwboth:
I am merging without review.

PR #141, opened 2023-02-01
Feature/image subtract

EStorvik:
Tried out operation overload to allow subtraction of DarSIA images. It is a small thing, and if used carelessly it can be a bad practice, but it is also convenient. What do you think?

Could also include more checks to check whether metadata coincide as well?

### DIFF
--- a/src/darsia/image/image.py
+++ b/src/darsia/image/image.py
@@ -240,8 +240,11 @@ class Image:
         """
         if self.img.shape != other.img.shape:
             warn("Images have different shapes. Resizing second argument to match.")
-            other_img = cv2.resize(other.img, self.img.shape[::-1][1:3])
-            return da.Image(self.img - other_img, copy.copy(self.metadata))
+            return da.Image(
+                self.img
+                - cv2.resize(np.copy(other.img), tuple(reversed(self.img.shape[:2]))),
+                copy.copy(self.metadata),
+            )
         else:
             return da.Image(self.img - other.img, copy.copy(self.metadata))
 
--- a/src/darsia/image/image.py
+++ b/src/darsia/image/image.py
@@ -228,7 +228,24 @@ class Image:
     def original_dtype(self) -> np.dtype:
         return self.metadata["original_dtype"]
 
-    # ! ---- Corresponding setter functions for metadata
+    # ! ---- Operation overloaders
+    def __sub__(self, other: da.Image):
+        """Subtract two images.
+
+        Arguments:
+            other (Image): image to subtract from self
+
+        Returns:
+            Image: difference image
+        """
+        if self.img.shape != other.img.shape:
+            warn("Images have different shapes. Resizing second argument to match.")
+            other_img = cv2.resize(other.img, self.img.shape[::-1][1:3])
+            return da.Image(self.img-other_img, copy.copy(self.metadata))
+        else:
+            return da.Image(self.img-other.img, copy.copy(self.metadata))
+
+
 
     def write(
         self,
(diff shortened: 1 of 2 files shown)


EStorvik:
And the copy of the regularization was an accident to include into this PR, but it has to be there anyways i think. Otherwise the input image is also modified, which is very annoying. I'm sure I fixed this more than 5 times before just that I forgot to merge my commit every time.

src/darsia/image/image.py
```diff
@@ -228,7 +228,24 @@ def timestamp(self, time: datetime) -> None:
...
+        Arguments:
+            other (Image): image to subtract from self
+
+        Returns:
+            Image: difference image
+        """
+        if self.img.shape != other.img.shape:
+            warn("Images have different shapes. Resizing second argument to match.")
+            other_img = cv2.resize(other.img, self.img.shape[::-1][1:3])
+            return da.Image(self.img-other_img, copy.copy(self.metadata))
```
jwboth: Small typo. Should be `other.img`

src/darsia/image/image.py
```diff
@@ -241,11 +241,9 @@ def __sub__(self, other: da.Image):
         if self.img.shape != other.img.shape:
             warn("Images have different shapes. Resizing second argument to match.")
             other_img = cv2.resize(other.img, self.img.shape[::-1][1:3])
-            return da.Image(self.img-other_img, copy.copy(self.metadata))
+            return da.Image(self.img - other_img, copy.copy(self.metadata))
```
jwboth, line 244: `other.img`

jwboth (approved):
Looks OK to me.

src/darsia/image/image.py
jwboth: ah no. The resize is wrong. It should reverse x and y. Or are you trying to do this  here? I am not following the index address. Using tuple(reversed(self.img.shape[:2])) usually worked out for me.

EStorvik: I think that the resize is correct. This was just the way I came up with on the fly to reverse x and y. At least it worked when I tested it. Had forgotten the tuple(reversed()) command.

Also I don't think it should be other.img. It is horrible notation from my side though. I'll fix it. And ill ad a copy so that the original other is not modified.

EStorvik: Check new update. I changed to the tuple(reversed()), it seems like it does exactly the same, but it reads way nicer, thanks!

src/darsia/image/image.py
```diff
@@ -240,8 +240,11 @@ def __sub__(self, other: da.Image):
         """
         if self.img.shape != other.img.shape:
             warn("Images have different shapes. Resizing second argument to match.")
-            other_img = cv2.resize(other.img, self.img.shape[::-1][1:3])
-            return da.Image(self.img - other_img, copy.copy(self.metadata))
+            return da.Image(
+                self.img
+                - cv2.resize(np.copy(other.img), tuple(reversed(self.img.shape[:2]))),
```
jwboth, line 245: Not sure whether the copy is needed here. I do not see how the resize would affect the image. But it is OK for me.

jwboth (commented):
We will have to rethink whether we will want to allow for images of differing size when considering 3d. Maybe it will be easier to just drop an error when compaitiboility is not ensured, and having an object or so which organizes the resizing (for 2d and 3d) instead of that such code will be included in all possible routines which need compatibility. To be revisited I would say.

jwboth:
For me it is good to go.

src/darsia/image/image.py
EStorvik: I don't know. I was just afraid that cv2's resize function modified the original image. But I checked now, and it doesnt. Ill remove the copy and merge then.

PR #162, opened 2023-04-19
Included nd regularization and solver functionality

EStorvik:
Several new files, and restructuring is necessary in the near future, but at least it runs.

Summary:
 - Moved updated derivatives (old ones are still there, but should be deleted, I did not in fear of ruining something old). I have also thought about a class structure for this, but I found that it worked nicely without, but I am open to changing this.
 - An array_slice function in utils. This one can be moved and redefined the places it is used, but I have found it very handy.
 - New solver class for heat-equation-like linear problem with three implementations:
       - Jacobi
       - Multigrid (V-cycle)
       - CG (basically a wrapper for SciPy's cg)
 - L2-regularization (comes for free through any of the linear solvers)
 - Anisotropic TVD with split Bregman algorithm. We now have MANY of these implemented. I think we should try to narrow it down and find a better structure for it. 
 - Provided one example for the 3D dicom image with both tvd and l2-regularization.


src/darsia/utils/linear_solvers/jacobi.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class Jacobi(da.Solver):
+    """
+    Jacobi solver for the problem: mass_coeff * x - diffusion_coeff * laplace(x) = rhs
+    """
+
+    def __init__(
+        self,
+        mass_coeff: Union[float, np.ndarray],
```
jwboth, line 20: Just to note, did you test this (and the other solvers) for array inputs? I cannot see any use of np.multiply or np.divide, which I would have expected instead of "/" for the latter.

src/darsia/utils/linear_solvers/mg.py
```diff
@@ -0,0 +1,235 @@
...
+from typing import Optional, Union
+
+import numpy as np
+
+import darsia as da
+
+
+class MG(da.Solver):
+    """
+    Muiltilvel solver for the problem: mass_coeff * x - diffusion_coeff * laplace(x) = rhs
```
jwboth, line 16: Any restrictions on the size of the input array? How do you take care of sizes which are not a power of 2? Maybe a very short comment here could be added.

src/darsia/utils/linear_solvers/mg.py
```diff
@@ -0,0 +1,235 @@
...
+            output [np.ndarray]
+
+        """
+
+        return self.mass_coeff * x - self.diffusion_coeff * da.laplace(
+            x, dim=self.dim, h=h
+        )
+
+    def restriction(self, x: np.ndarray) -> np.ndarray:
+        """Restrict x, i.e., coarsen it by a factor 2. Even and odd indices are averaged.
```
jwboth, line 83: What if `x` does not have an even size?

src/darsia/utils/linear_solvers/solver.py
```diff
@@ -0,0 +1,30 @@
+from typing import Optional, Union
+
+import numpy as np
+
+
+class Solver:
```
jwboth, line 6: All names (Solver, CG, MG, Jacobi) sound very generic though have a very tailored use case in mind. OK for now. We will see whether we want to change names at some point.

src/darsia/utils/linear_solvers/solver.py
```diff
@@ -0,0 +1,30 @@
...
+        self.mass_coeff = mass_coeff
+        self.diffusion_coeff = diffusion_coeff
+        self.dim = dim
+        self.maxiter = maxiter
+        self.tol = tol
+        self.verbose = verbose
+
+    def __call__(self, x0: np.ndarray, rhs: np.ndarray) -> np.ndarray:
+        print("Provide a call ,method for the solver.")
+        return x0
```
jwboth, line 30: This looks like you wish to make the method abstract.
```suggestion
    @abc.abstractmethod
    def __call__(self, x0: np.ndarray, rhs: np.ndarray) -> np.ndarray:
        pass
```

src/darsia/utils/linear_solvers/solver.py
```diff
@@ -0,0 +1,30 @@
+from typing import Optional, Union
```
jwboth, line 1:
```suggestion
import abc
from typing import Optional, Union
```
see below

jwboth (approved):
Looks very good! Thanks! I just made very minor comments and suggestions. Feel free to accept or not. And feel free to merge whenever you feel ready.

src/darsia/utils/array_slice.py
```diff
@@ -0,0 +1,33 @@
...
+from typing import Union
+
+import numpy as np
+
+
+def array_slice(
+    a: np.ndarray,
+    axis: int,
+    start: Union[int, None],
+    end: Union[int, None],
```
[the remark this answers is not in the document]
EStorvik: Oops, you are of course right!

src/darsia/restoration/split_bregman_tvd.py
```diff
@@ -0,0 +1,131 @@
...
+            nrm = np.linalg.norm(im_new - im) / im0_nrm
+            if verbose:
+                print(f"The relative increment at iteration {i} is {nrm}")
+            if nrm < tol:
+                break
+        im = im_new
+        if verbose:
+            print(f"The the energy functional iteration {i} is {_functional(im)}")
+
+    # Might need to put a clip here to avoid numerical issues
```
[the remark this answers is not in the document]
EStorvik: I haven't seen this problem since I fixed the last bug so perhaps it will never happen... In theory I guess it should not.

src/darsia/utils/linear_solvers/mg.py
EStorvik: There are no restrictions in princliple, but my method might not be very good.. Now I'm just disregarding the last row/column when this is the case..

src/darsia/utils/linear_solvers/solver.py
EStorvik: I thought about it, but I thought that should be for when we actually want to solve other linear systems. As of now I have not come up with any.

src/darsia/utils/linear_solvers/mg.py
EStorvik: Then we lose the last row/column/..

src/darsia/utils/linear_solvers/solver.py
EStorvik: haha, yes! thanks!

src/darsia/utils/linear_solvers/mg.py
jwboth: OK. Should be OK. Maybe add a comment explicitly stating this?

src/darsia/utils/linear_solvers/mg.py
jwboth: OK. Does in this case the prolongation add this row again, or does it always double?

EStorvik: It actually always doubles, although it should have added the row again. I totally forgot about this. I will look at it again.

EStorvik: Im just merging this with a note that it should be improved upon in the future.

src/darsia/utils/linear_solvers/jacobi.py
EStorvik: I tested it all for array inputs, and im like 99% sure that np.multiply is equivalent with * (similar with / and divide)

src/darsia/utils/linear_solvers/jacobi.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def diag(self, h: float = 1) -> Union[float, np.ndarray]:
+        """
+        Compute diagonal of the Jacobi matrix with a given mesh diameter h
+
+        args:
+            h (float): mesh diameter
+
+        Returns:
+            Union[float, np.ndarray]: diagonal of the Jacobi matrix
```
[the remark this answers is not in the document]
EStorvik: It should have the same dimensions as the diffusion_coefficient and mass_coefficient matrices (which would be same as original monochromatic image, or just stored as a float if neither are spatially varying).

src/darsia/utils/linear_solvers/jacobi.py
EStorvik: It might be that I have misunderstood though..?

src/darsia/utils/linear_solvers/cg.py
```diff
@@ -0,0 +1,88 @@
...
+import scipy.sparse as sps
+from scipy.sparse.linalg import LinearOperator
+
+import darsia as da
+
+
+class CG(da.Solver):
+
+    """Conjugate gradient solver based on scipy's cg.
+    For the problem: mass_coeff * x - diffusion_coeff * laplace(x) = rhs
```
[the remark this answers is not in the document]
EStorvik: Yes.. I'll look into this. I think I disregarded it since it mostly is just the ell-term entring there in the TVD-case.

src/darsia/utils/linear_solvers/jacobi.py
jwboth: I think you should be right. Disregard my comment.

src/darsia/utils/linear_solvers/cg.py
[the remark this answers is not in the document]
EStorvik: I have now modified this in CG, but not in the Jacobi where there is some more structure is needed. Ill make this into a new PR for later.

PR #183, opened 2023-05-09
WIP: Establish docs infrastructure

jwboth:
The goal is to provide a better first layer of a documentation, i.e., on rst level. Future maintanence will include improving docstrings.


src/darsia/presets/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Presets for workflows to study concentration dynamics in heterogeneous media.
+
+"""
```
jwboth:
```suggestion
"""Presets for workflows to study concentration dynamics in heterogeneous media."""
```

src/darsia/signals/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Dimensionsional reduction and transformation of signals.
+
+"""
```
jwboth:
```suggestion
"""Dimensionsional reduction and transformation of signals."""
```

src/darsia/utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Utility functionalities.
+
+"""
```
jwboth:
```suggestion
"""Utility functionalities."""
```

PR #206, opened 2023-07-04
Clean up regularization

jwboth:
Unify TVD (and L2 regularization). Modify signatures, add support for `darsia.Image` objects, and provide wrapper combining the new methods with skimage methods. Showcase the new methods in an updated example for 3d data.


jwboth:
@EStorvik I have made some modifications to the signature and structure of the code to allow for a more unified presentation. It would be good if you review these changes at some point. I am merging to main already to continue other developments.

PR #213, opened 2023-07-05
Rename l2 regularization -> h1 regularization.

jwboth:
The previsouly called L2 regularization is more often related to as H1 regularization, as it is the H1 seminorm which is added to the data term. Thus, the regularization is characterized by the diffusion equation (compared e.g. to TV regularization where the TV term is added).


jwboth:
@EStorvik FYI

PR #214, opened 2023-07-14
Anisotropic regularization parameter

EStorvik:
Added functionality to take physical derivatives and moved also included in the jacobi and cg solvers. However, it seemed to work poorly when tested with the tvd. It might be because the scaling in the linear system that is solved for the tvd gets off when the laplace term suddenly gets huge compared to the mass term. I guess that there in a heat equation setting normally would be some sort of cfl condition for the step size parameter here (but we don't have that). It could also be that I made a mistake somewhere and therefore it did not work. Nevertheless, I don't have more time to work on this for the coming weeks, so I removed it from the TVD and made a PR with the other changes. 

It is now possible to prescribe different regularization parameters for the different directions when doing tvd (see example). 

Sidenote: I do wonder if my placement of mu in the tvd-algorithm is suboptimal, and makes things converge slower than necessary. What I mean is that I could have substituted \mu\nabla g instead of just \nabla g when splitting the minimization problem. I'll have this in the back of my head and perhaps I'll try the other way arond later.

### DIFF
--- a/examples/regularization.py
+++ b/examples/regularization.py
@@ -6,31 +6,27 @@ import numpy as np
 import darsia
 
 # Load 3D image
-# image_folder = f"{os.path.dirname(__file__)}/images/"
-image_folder = "C:/Users/erst/src/image_analysis/images/"
+image_folder = f"{os.path.dirname(__file__)}/images/"
 img_path = image_folder + "dicom_3d.npy"
 img = darsia.imread(img_path, dimensions=[0.1, 0.1, 0.1], dim=3)
 
 print("Image shape: ", img.img.shape)
 
-# Make regularization parameter heterogenous (for illustration purposes)
-mu_physical = 4.7e-4 * np.ones_like(img.img)
-mu_physical[:, 0:100, :] = 0.5 * 4.7e-4
-
-mu_physical = 4.7e-4
-
+# Make regularization parameter heterogenous and anisotropic (for illustration purposes)
 mu = np.ones_like(img.img)
 mu[:, 0:100, :] = 0.5
 
+mu_anisotropic = [10 * mu, mu, mu]
+
 # Regularize image using anisotropic tvd
-img_regularized_physical_tvd = darsia.tvd(
+img_regularized_anisotropic_tvd = darsia.tvd(
     img=img,
     method="heterogeneous bregman",
     isotropic=False,
-    weight=mu_physical,
+    weight=mu_anisotropic,
     omega=0.1,
     max_num_iter=30,
-    eps=1e-4,
+    eps=1e-6,
     dim=3,
     verbose=True,
     solver=darsia.Jacobi(maxiter=20),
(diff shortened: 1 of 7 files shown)


examples/regularization.py
```diff
@@ -8,25 +8,43 @@
...
+    weight=mu_anisotropic,
+    omega=0.1,
+    max_num_iter=30,
+    eps=1e-6,
+    dim=3,
+    verbose=True,
+    solver=darsia.Jacobi(maxiter=20),
+    # solver=darsia.CG(maxiter=20, tol=1e-3),
+)
+
```
jwboth, line 35: Add comment like # Regularize image using isotropic tvd

src/darsia/restoration/split_bregman_tvd.py
```diff
@@ -48,16 +50,26 @@ def split_bregman_tvd(
...
     # Store input image and its norm for convergence check
     img_nrm = np.linalg.norm(img)
 
+    # Controll that mu has correct length if its a list
+    if isinstance(mu, list):
+        assert len(mu) == dim, "mu must be a list of length dim"
+
     # Feed the solver with parameters, follow the suggestion to use double the weight
-    # for the diffusion coefficient if no value is provided.
+    # for the mass coefficient if no value is provided.
```
jwboth, line 65: I would suggest to enhance the comment to explain why additionally ell is set to 1 if it is zero. Furthermore, I would use an inexact check whether ell is (almost) zero.

src/darsia/utils/linear_solvers/jacobi.py
```diff
@@ -21,7 +21,9 @@ class Jacobi(da.Solver):
 
     """
 
-    def _neighbor_accumulation(self, im: np.ndarray) -> np.ndarray:
+    def _neighbor_accumulation(
+        self, im: np.ndarray, h: Optional[Union[float, list]] = None
+    ) -> np.ndarray:
         """Accumulation of neighbor pixels.
```
jwboth, line 27: Better documentation would be good here. What is "accumulation" and how does it depend on the grid size provided?

src/darsia/utils/linear_solvers/jacobi.py
```diff
@@ -37,44 +39,72 @@ def _neighbor_accumulation(self, im: np.ndarray) -> np.ndarray:
...
+        if h is None:
+            voxel_weights = self.dim
+        else:
+            if isinstance(h, float):
+                voxel_weights = self.dim / (h**2)
+            else:
+                assert len(h) == self.dim, "h must be a float or a list of length dim"
+                voxel_weights = sum([1 / h[i] ** 2 for i in range(self.dim)])
+
+        return self.mass_coeff + 2 * self.diffusion_coeff * voxel_weights
```
jwboth, line 94: Should the 2 be moved into the definition of voxel_weights in line 89?

jwboth (approved):
Looks OK in principle. Is the dependence on the mesh size actually by default used in the TVD algorithm? Or would one have to manipulate the code to make things work?

Your observations regarding convergence and performance issues are of course not something we should ignore. Just to understand: Is the current implementation working as before? Or would I have to tune all my configurations for previous analyses, e.g., the benchmark analysis?

Interesting side note.

Issue #265, opened 2024-01-11
Object oriented definition of coordinates, voxels, matrix indexing etc.


jwboth:
Currently, points in the image can be specified throughtheir coordinate or voxel location. In addition, voxel coordinates (in particular in 2d) are used in different formats (matrix indexing, cartesian indexing, etc.). As external tools as opencv (in particular) uses different formats, making an internally consisten choice for DarSIA is difficult. Thus, we need to stick to the different formats. To remove confusion, the aim however should be to use objects which uniquely identify the format, i.e., we need to define objects `class Coordinate` (Cartesian coordinate), `Voxel` (voxel coordinate, which can have different indexing modes and allows for transforming the indexing through helper routines).

jwboth:
Still various transformation objects do not make use of `Voxel`, `Coordinate` etc. A unified use is envisioned.

PR #266, opened 2024-01-13
Voxels, Coordinates, and Generalized perspective transform

jwboth:
A native implementation of a general perspective transform is provided whcih concatenates standard perspective transforms, bulge and stretch operations. These are the standard routines entering `CurvatureCorrection.` 

The implementation is based on a newly introduced concept of `Voxel` and `Coordinate` as well as `VoxelArray` and `CoordinateArray`. These are subclasses of `np.ndarray` but explicitly carry the information of how to understand these. In addition, the types are used in the transformation routines of `CoordinateSystem`. This in part addresses Issue #265 . 

The graphical assistants also make use of the new point characterizations. 

However, not alle objects employ the new objects instead of standard numpy arrays.

### DIFF
--- a/src/darsia/__init__.py
+++ b/src/darsia/__init__.py
@@ -14,6 +14,7 @@ from darsia.measure.integration import *
 from darsia.measure.emd import *
 from darsia.measure.wasserstein import *
 from darsia.utils.box import *
+from darsia.utils.point import *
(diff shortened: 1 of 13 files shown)


jwboth:
For completeness, the test coverage and documentation coverage for the introduced functionality is satisfactory.

PR #269, opened 2024-01-14
Use mypy as part of the github workflow 

jwboth:
The plan is to use this PR for updating the code base to be inline with mypy.


jwboth:
Just a start... more work to come. Mypy deactivated from github actions again to ensure positive checks for PRs.

PR #274, opened 2024-01-16
Base transformation

jwboth:
Introduce a base transformation as super class for geometric shape transformation. Add corresponding correction object, which is forms a basis purely based on shape transformations. 

This PR comes with minor enhancements of other pieces of code.


jwboth:
Tests added covering in part the new (and in part old) functionality. Less rigorous testing of the generalized pespective map.

PR #298, opened 2024-02-08
BUG: Updated Image initialization and assembly order in assemble

EStorvik:
Updates were made to the class that had not been reflected in the assemble function, and as a result it did not work. It should work now. I do however think that the blend and assemble does not work.

### DIFF
--- a/src/darsia/image/patches.py
+++ b/src/darsia/image/patches.py
@@ -515,18 +515,18 @@ class Patches:
         assembled_img = np.zeros(
             (0, *self.base.img.shape[1:]), dtype=self.base.img.dtype
         )
+        print(assembled_img)
 
         # Create "image-strips" that are assembled by concatenation
-        for j in range(self.num_patches[1]):
+        for j in range(self.num_patches[0]):
 
             # Initialize the row with y coordinate j of the final image
             # with the first patch image.
             rel_roi = self.relative_rois_without_overlap[j][0]
             assembled_y_j = self.patches[j][0].img[rel_roi]
-
             # And assemble the remainder of the row by concatenation
             # over the patches in x-direction with same y coordinate (=j).
-            for i in range(1, self.num_patches[0]):
+            for i in range(1, self.num_patches[1]):
                 rel_roi = self.relative_rois_without_overlap[j][i]
                 assembled_y_j = np.hstack(
                     (assembled_y_j, self.patches[j][i].img[rel_roi])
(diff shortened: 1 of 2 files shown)


src/darsia/image/patches.py
```diff
@@ -638,9 +636,8 @@ def blend_and_assemble(self, update_img: bool = False) -> darsia.Image:
         da_assembled_img = darsia.Image(
             img=assembled_img,
             origin=self.base.origin,
-            width=self.base.width,
-            height=self.base.height,
-            color_space=self.base.colorspace,
+            dimensions=self.base.dimensions,
+            color_space=self.base.color_space,
```
jwboth, line 640: This is not general. `Image` in general does not have color space e.g. In general one can just use `metadata()` in the `kwargs` of `Image`. Also instead of `Image`, I guess using `type` of the input will be most general to ensure the output type is identical to the input type.

src/darsia/image/patches.py
```diff
@@ -543,9 +542,8 @@ def assemble(self, update_img: bool = False) -> darsia.Image:
...
```
jwboth, line 546: See other comment on generality.

jwboth (changes requested):
Thanks for taking care of that! This has (in part) not been spotted before, as the test coverage is not sufficient. I suggest to add very simple tests. If you wish, I can take care of that - feel free to go ahead if you feel like it!

src/darsia/image/patches.py
EStorvik: I think I updated this correctly now with the metadata (please have a look). I was however a bit unsure of what you mean by type. I'm not 100% sure how it works. Can you either explain quickly or show me the correct way (it sounds like its supposed to be easy and allow for more flexibility?).

EStorvik:
I also added a very simple test. Is this something of the sort you had in mind?

src/darsia/image/patches.py
```diff
@@ -637,10 +633,7 @@ def blend_and_assemble(self, update_img: bool = False) -> darsia.Image:
         # Define resulting darsia image
         da_assembled_img = darsia.Image(
```
jwboth, line 634:
```suggestion
        da_assembled_img = type(self.base)(
```
I would also suggest to return the same type as the original image. So if an `OpticalImage` e.g. is provided as input (which knows about `color_space` while the others do not), also the assembled image does. Could you include this change and similarly for the other creations of Images in the same file? Then this bug would be also corrected at the same time.

jwboth (approved):
Looks good, including the test! If you make the last changes, feel free to merge. Thanks!

src/darsia/image/patches.py
EStorvik: This makes sense. Haven't thought of using it this way before... Very neat. I'll update it.

PR #365, opened 2024-08-19
W1 distance with anisotropic weights

jwboth:
Extend W1 distance computations to allow for anisotropic cell-based weights.


jwboth:
Obsolete through #366

jwboth:
#366

PR #366, opened 2024-08-21
W1 with anisotropic weights

jwboth:
Use correct harmonic averaging in consistent manner. Apply to anisotropic weights. Only cell based averaging.


jwboth:
Ignore the failing W1 tests for now as these are all based on face based weight reconstruction - which is outdated.

PR #379, opened 2024-09-27
Improvements in petsc usage

enricofacca:
- The formulation "flux_reduced" for the solution of linear systems in  the wasserstein.py script is now covered with ksp solvers.

- Tuning of hypre preconditioner is added for 3d case.

- We can now pass extra ksp option using the dict. "petsc_options" in "linear_solver_options":  .

- Using petsc matrices and routined the assembly of the schur complement does not required to convert from scipy to petsc several time.

### DIFF
--- a/src/darsia/measure/wasserstein.py
+++ b/src/darsia/measure/wasserstein.py
@@ -15,6 +15,7 @@ import numpy as np
 import pyamg
 import scipy.sparse as sps
 from scipy.stats import hmean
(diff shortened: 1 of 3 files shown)


enricofacca:
I included some functions to buffer the assembling phase of the schur complement for non PETSCs solvers.

enricofacca:
Is it now possible to merge the code?

PR #393, opened 2025-05-06
New method for Wasserstein 1. Small bug fund. 

enricofacca:
Included the approach described in [this paper ](https://epubs.siam.org/doi/10.1137/18M118640X) .
It seems less accurate but it base entirely  on pure Poisson problem.

I think a found a bug in test/unit/test_wasserstein.py. I was not running any test, since solvers's list was empty.

This lead to a revaluation on how to compute is convergence is a achieved. I replaced the residuum 
q/|q| - \nabla u =0
with
q - |q|\nabla u = 0
in order to measure differently the residuum on the region were |q| (the transport density is small).


src/darsia/measure/wasserstein.py
```diff
@@ -1587,6 +1601,17 @@ def _solve(self, flat_mass_diff: np.ndarray) -> tuple[float, np.ndarray, dict]:
...
+                transport_density_faces = self.transport_density_faces(flux)
+                res_opt = self.mass_matrix_faces.dot(
+                    solution_i[self.flux_slice]
+                ) - transport_density_faces * self.div.T.dot(
+                    solution_i[self.pressure_slice]
+                )
+
+                residual_opt = np.linalg.norm(res_opt, 2) / old_distance
+                residual_div = np.linalg.norm(res_pressure, 2)
+                res_pde = np.sqrt(residual_opt**2 + residual_div**2)
```
jwboth, line 1614: Do I understand correctly that the extra calculation are just about how the residual norm is calculated for the convergence check but the computations of the next iterate remain the same?

src/darsia/measure/wasserstein.py
```diff
@@ -2090,6 +2117,686 @@ def _solve(self, flat_mass_diff: np.ndarray) -> tuple[float, np.ndarray, dict]:
         return new_distance, solution_i, info
 
 
+class WassersteinDistanceGproxPGHD(darsia.EMD):
```
jwboth, line 2120: I did not go into the details and trust you that the method is doing what is intended, seeing you added all required tests.

jwboth (approved):
Looks good to me. Am just interested in the curiosity, I posted. Thanks a lot!

PR #395, opened 2025-05-28
Add OpticalImage.encode() method

TheBB:
@jwboth

This lets us write an image to bytes without having to go via a file on disk.


jwboth (approved):
Thanks a lot, @TheBB !

PR #401, opened 2025-07-20
Run pytest in parallel

jwboth:
Use pytest-xdist to run tests in parallel.


jwboth:
Note: Current speed up minimal. About 20 seconds.

PR #406, opened 2025-11-02
Updated requirements.txt

Teggo19:
Removed version specification for matplotlib, and added shapely and openpyxl to requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 numpy
 opencv-python
 scipy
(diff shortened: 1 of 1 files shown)


jwboth:
Looks good, thanks @Teggo19 !

Could you add a brief description of the changes at the top of the PR. Such a description typically does not need to be long, depending on the amount of changes proposed. It is just used for better tracking of the different PRs.

PR #408, opened 2026-01-22
Update wasserstein tests

Teggo19:
Created tests of the Wasserstein Distance.
One test is rectangular blocks at different angles compared with the analytical solution.
The second test is rectangular blocks with a wall placed in between the blocks, forcing the mass to be transported around the corner.


tests/unit/test_wasserstein_error.py
```diff
@@ -0,0 +1,165 @@
...
+    "regularization": regularization,
+    "scaling": scaling,
+    "depth": 0,
+    "verbose": False,
+    "return_info": True
+}
+results = np.array([[ 0.00416437,  0.0039679 ],
+ [ 0.00474455,  0.00259308],
+ [ 0.0048696,   0.00369112],
+ [ 0.00416437,  0.0039679 ],])
```
jwboth: Better defined inside the relevant test. If used across multiple tests, use some common setup routine or fixture.

src/darsia/utils/wasserstein_analytical_sol_block.py
```diff
@@ -0,0 +1,3 @@
+# Formatted code in wasserstein_analytical_sol_block.py using Black and isort
+
+# Your auto-formatted content goes here...
```
jwboth: Copilot did manage to include some errors ... Could you locally run black and isort on the documents? And remove the comments again. Maybe it is worth checking what the copilot commit introduced here - sorry for that! Possibly just revert the last comment.

PR #615, opened 2026-07-15
Gui dev

Teggo19:
The creation of a new gui that helps the user adjust the necessary settings for each action.

Still need to fill out the settings_mapping.toml file appropriately to make the GUI usable with the current workflows. There is also a need to add the workflows and correct tags.


src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
+from PySide6.QtWidgets import (QScrollArea, QMainWindow, QApplication, QLabel, QWidget, QVBoxLayout, QPushButton,
```
jwboth: Do a minimal effort adding some docstrings for the module, the main class, and methods.

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+import toml
+import ast
+import os
+import subprocess
+
+
+class MainWindow(QMainWindow):
+    def __init__(self):
+        super().__init__()
+        self.setWindowTitle("Darsia")
```
jwboth:
```suggestion
        self.setWindowTitle("DarSIA")
```

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+        tabs.addTab(calibration_container, "Calibration")
+        tabs.addTab(analysis_container, "Analysis")
+        upper_mid_layout.addWidget(tabs)
+
+        # Creating the setup-tab
+        setup_items = [
+            ("All", "all"),
+            ("Depth", "depth"),
+            ("Segmentation", "segmentation"),
+            ("Facies", "facies"),
```
jwboth, line 86: We can keep them for now. Just FYI: For the standard user, only the options "all", "protocol", and "show plots" are used, I would say.

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+        # Create central widget with all components
+        main_container = QWidget()
+        main_layout = QVBoxLayout(main_container)
+        self.setCentralWidget(main_container)
+        main_layout.addWidget(upper_container)
+        main_layout.addWidget(content_splitter)
+
+    def load_settings_mapping(self):
+        """Load the settings_mapping.toml file."""
+        mapping_file = os.path.join(os.path.dirname(__file__), "settings_mapping.toml")
```
jwboth: I like `from pathlib import Path` more than `os.path`, and it is consistently used throughout DarSIA. Would it be possible to change that?

src/darsia/gui/settings_mapping.toml
```diff
@@ -0,0 +1,16 @@
+
```
jwboth, line 1: OK for now. I wonder whether we will want to organize one gigantic file in the end, or split them into logical subfiles. I also have an idea about further "info" for each setting with some helper text, that can be displayed in the gui when clicking on a "?" button or so (TBC). This will make the file even longer.

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+def set_dict_value(inp_dict, path, value):
+    local_dict = inp_dict
+    for key in path[:-1]:
+        local_dict = local_dict[key]
+    local_dict[path[-1]] = value
+
+if __name__ == '__main__':
+
+    #with open("test.toml", "r") as f:
+    #    config = toml.load(f)
```
jwboth: Do we need the test.toml file? If it is more convenient for you to keep it as an example file and to develop, let's keep it. But maybe then add a TODO comment for these lines and mentioning a potential removal of test.toml.

In the future, an example.toml file, that is explained in a tutorial, would be a natural thing to do.

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+        return setting_container, file_edits
+
+
+def set_dict_value(inp_dict, path, value):
+    local_dict = inp_dict
+    for key in path[:-1]:
+        local_dict = local_dict[key]
+    local_dict[path[-1]] = value
+
+if __name__ == '__main__':
```
jwboth: My sense is, we may want to be even more granular here and introduce two files. One `gui.py` that just containts the `__main__` script and mainly serves as entry point, and one that designs the gui and defines the `MainWindow` class. What about introducing that split already now? Or is there any reason for not doing that?

src/darsia/gui/test.toml
```diff
@@ -0,0 +1,11 @@
+[rig.test]
```
jwboth: I assume this file can be removed (now or later)?

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+
+    def analysis(self):
+        """Analysis action: get checked specifics and display relevant settings."""
+        checked_ids = self.get_checked_checkbox_ids(self.analysis_checkboxes)
+        self.display_settings("analysis", checked_ids)
+        self.print_log(f"Analysis: Checked specifics - {checked_ids}")
+
+    def run_setup(self):
+        #subprocess.run(["python", "setup_protocol"])
+        pass
```
jwboth: FYI: I am expecting here we will need a try-except statement, that catches expressions and prints them to screen or so, if an error occurs. Here we will actually need to work out a better way of informing the user when and what goes wrong. TBC

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+
+    def load_settings_mapping(self):
+        """Load the settings_mapping.toml file."""
+        mapping_file = os.path.join(os.path.dirname(__file__), "settings_mapping.toml")
+        if os.path.exists(mapping_file):
+            return toml.load(mapping_file)
+        return {}
+
+
+    def save_settings(self):
```
jwboth, line 214: FYI: `save_settings` should also be executed when clicking `ctrl+s`. Many users expect that.

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+        self.chosen_files[key] = {"path": "", "label": path_label, "is_directory": is_directory, "filter": file_filter}
+        
+        # Connect button to file dialog
+        browse_button.clicked.connect(lambda: self.browse_file(key))
+        
+        chooser_layout.addWidget(browse_button)
+        chooser_layout.addWidget(path_label)
+        chooser_layout.addStretch()
+        return chooser_container, path_label
+        #parent_layout.addWidget(chooser_container)
```
jwboth: can be removed?

src/darsia/gui/gui_main.py
```diff
@@ -0,0 +1,675 @@
...
+                               QHBoxLayout, QTabWidget, QSplitter, QCheckBox, QFileDialog)
+from PySide6.QtCore import Qt
+import toml
+import ast
+import os
+import subprocess
+
+
+class MainWindow(QMainWindow):
+    def __init__(self):
```
jwboth, line 27: OK for now, but should be organized in shorter methods in near future. Do not prioritize that for now.

jwboth (approved):
Super cool! Looking forward to switch to this soon! Just some minimal comments, with only a few todos. Most other comments are only FYI type comments. Just resolve them as you like, and merge PR once you are ready.

jwboth:
@Teggo19 I also requested a review by Copilot. Let's see how useful that is. If not meaningful, just ignore and resolve the comments.

Teggo19:
Todo:
- ctrl+s for save
- shorter methods

PR #659, opened 2026-08-17
Patchwise illumination correction

thibault-vautier:
updated code


jwboth (approved):
Will do the static update in a later PR.

